Issue #4, opened 2018-09-11
"Open in Text Editor" button


huderlem:
AMap does this with XSE. Need to be able to open up a relevent script file with text editor with just a single click. Let the OS handle opening it up.

(originally from https://github.com/yenatch/pretmap/issues/60)

huderlem:
aaaaaa123456789 commented on Jul 16

A typical strategy is to just use a substitution character for the filename in the command line — so you can configure it to something like sillyeditor --file $ --some-other-option 1 and the program will know to replace the $ with the quoted filename.

PR #24, opened 2018-09-12
"Open Map Scripts" button

garakmon:
(copy of [https://github.com/yenatch/pretmap/pull/70](https://github.com/yenatch/pretmap/pull/70))

Adds a button to open a map's scripts in a the system default text editor.

Should close Issue #4 

demo:
[Alt text](https://www.youtube.com/watch?v=qFW7L9HFJNg)

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -20,6 +20,7 @@
 #include <QDialogButtonBox>
 #include <QProcess>
 #include <QSysInfo>
+#include <QDesktopServices>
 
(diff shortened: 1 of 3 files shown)


huderlem:
Rather than using those different commands for different OS's, we should use Qt's built-in functionality:
```
QString path = QDir::cleanPath(editor->project->root + QDir::separator() + "data/maps/" + editor->map->name + "/scripts.inc");
QDesktopServices::openUrl(QUrl(path));
```

Plus, the Windows command didn't work for me.

PR #25, opened 2018-09-12
Cursor update

garakmon:
(copy of [https://github.com/yenatch/pretmap/pull/69](https://github.com/yenatch/pretmap/pull/69))

Adds:
- toggleable move / drag mode with cursor
- changing cursors based on selected edit mode


garakmon:
@huderlem  I did not update the way the different cursors are displayed over the map--was not sure if there was a consensus on this. Also now that you have write access you can push your changes to my branch or whatever.

huderlem:
I think the cursors I demo'd are a bit better, so I can push that.  In the meantime, would you mind adding the ability to toggle this cursor behavior on and off, like Deokishisu suggested?  Probably just a top dropdown menu option.

garakmon:
yep. I'll add it under "View"

garakmon:
I added a checkable menu item under View>Better Cursors (what do you think of this name?). I also made the user's last set cursor style saved in the settings.

PR #38, opened 2018-09-19
Use project values for tilesets

Diegoisawesome:


tileset.cpp
```diff
@@ -9,6 +9,13 @@ Tileset::Tileset()
 
 }
 
+int Tileset::num_tiles_primary;
```
huderlem: I'd prefer is these were non-static fields in `Project`, instead.  Will help a bit with a refactor in the future.

Diegoisawesome: Okay, so how should I access them from the Map and Tileset classes? Just pass them as parameters all the way down to where they're needed?

Issue #37, opened 2018-09-18
Zooming / Moving Map Bug


Dangie:
Step 1: Zoom a few times
Step 2: Change current map to another map

Now the map is unusable and the move tool is not working.
Zooming is now broken at this point 

Switching back to the first map and zooming in or out has fixed it for me.
Relaunching also works too..

Here are some screen shots:

Editor after project is loaded:
![editor_normal](https://user-images.githubusercontent.com/29085916/45709622-c4d54f80-bbc7-11e8-9ee8-c70962747fd3.png)

Bugs:
![editor_bug](https://user-images.githubusercontent.com/29085916/45709618-c3a42280-bbc7-11e8-9d1f-fdb76c4cd20b.png)

![editor_bug2](https://user-images.githubusercontent.com/29085916/45709620-c43cb900-bbc7-11e8-9e79-3facb790c7e5.png)

![editor_bug3](https://user-images.githubusercontent.com/29085916/45709621-c43cb900-bbc7-11e8-8b6e-6ad1a068cead.png)

Issue #35, opened 2018-09-16
Changing the file path of the project then reopening porymap it still thinks the project is there.


ProjectRevoTPP:
1] open pokeruby
2] close porymap
3] rename pokeruby folder to pokeruby2 or something
4] open porymap again
5] buggy project open that doesnt exist. it should raise an error and close the project internally.

huderlem:
It no longer crashes, and will be a blank slate as if you are opening porymap for the first time.

Issue #34, opened 2018-09-16
File > Close Project


ProjectRevoTPP:

garakmon:
Is the idea here that closing a project will close the application, and when opening the app again will not look for the project at that path? Or is the idea more like it will un-load the current project in the editor and it will be like there is no project open (how it is on first use), and if you close from this state it will not try to reopen last project folder?
I hope that makes sense.

garakmon:
c0674d96769e62d73181d717c5b76fc931612841

Issue #28, opened 2018-09-14
Support smart paths with interior corners


Deokishisu:
- [ ] Support for FR-style paths in smart pathing. 

<img width="182" height="139" alt="Image" src="https://github.com/user-attachments/assets/5342bffc-c27d-4c6a-82e2-83ac4291d304" />

is an example usage. They are also used in Slateport City in RSE, with corner tiles when the path changes direction.
- [x] Opening map scripts could open right to the line of the open entity's script. Probably could wait for a standard pret script editor.
- [x] It seems that mouseclicks to copy, draw, etc. are disabled with any tool but the pencil manually selected. Not sure if that's intended behavior. Weird to have my shortcuts not work because I'm in eyedropper mode.
- [x] Add Ctrl+Mousewheel Scroll as a shortcut to zoom in or out on a map, like standard browsers.
- [x] When selecting where a warp would lead to, perhaps separate the maps by bank and allow those banks to be expanded? There's a LOT of scrolling involved in getting warps correct.
- [x] Perhaps show a movement "radius" for NPCs that can move to see how their pathing is. Like, select an NPC and tiles that they can move on with their current movement type are highlighted.

BigBahss:
The second checkbox can be marked with #325 merged.

GriffinRichards:
Updating the title to reflect the remaining point here (smart paths with interior corners).

The other points should have other things to address them:
- No particular shortcut requests as discussed (eyedropper is a subset of the pencil tool anyway)
- For warps you can start typing the name of a map and it'll autocomplete, rather than scrolling through the list. I think a sub-menu with collapsible sections would be overcomplicating it
- I'd like to do some kind of path highlighting, but there's far too much special handling in the movement code for Porymap to understand. This is a problem for the events API (in progress already, I have a prototype path highlighter below):

<img width="750" height="892" alt="Image" src="https://github.com/user-attachments/assets/0538691a-a66c-4d83-8df7-f8f06cca5e09" />

![Image](https://github.com/user-attachments/assets/dc05540c-f4e5-465e-b11f-a1dcb439246d)

PR #39, opened 2018-09-20
Bugs 1

garakmon:


garakmon:
Okay I think I am done with this. I: Fixed the bug in Issue #37, added scrolling the mapList to highlight the current map (#26 ), added some version of a "Close Project" function (#34) that solves #35), added CTRL + mousewheel scroll as suggested in #28.

Issue #2, opened 2018-09-11
A document providing an install guide


huderlem:
(originally from https://github.com/yenatch/pretmap/issues/67)

PR #48, opened 2018-10-08
Add installation instructions and fix compile issues

phase:
By changing the call to QString#chopped to QString#left, we remove the
Qt 10 dependency.

I also changed the C++ version to C++11 as that's what the code is
using my compiler (and others) didn't default to it.

Ref #2

### DIFF
--- a/src/core/parseutil.cpp
+++ b/src/core/parseutil.cpp
@@ -77,7 +77,7 @@ QList<HealLocation>* ParseUtil::parseHealLocs(QString text) {
     int i = 1;
     for (auto line : lines){
         if (line.contains("MAP_GROUP")){
-            QList<QString> li = line.replace(" ","").chopped(2).remove('{').split(',');
+            QList<QString> li = line.remove(QRegExp("[{|} ]")).split(",");
             HealLocation hloc = HealLocation(li[1].remove("MAP_NUM(").remove(")"), i, li[2].toUShort(), li[3].toUShort());
             parsed->append(hloc);
             i++;
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@ porymap.pro.user
 moc_*
 qrc_*
 porymap.app*
+porymap
 Makefile
-ui_mainwindow.h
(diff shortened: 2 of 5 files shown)


src/core/parseutil.cpp
```diff
@@ -77,7 +77,7 @@ QList<HealLocation>* ParseUtil::parseHealLocs(QString text) {
     int i = 1;
     for (auto line : lines){
         if (line.contains("MAP_GROUP")){
-            QList<QString> li = line.replace(" ","").chopped(2).remove('{').split(',');
```
huderlem: `chopped()` and `left()` do different things, so I doubt this code still behaves correctly.

garakmon: I'd be surprised if the app even opened

phase: Ah, I read the documentation wrong. I'll fix this up.

garakmon: this line should work if it's really an issue `QList<QString> li = line.remove(QRegExp("[\{|\}]")).remove(" ").split(",");`

phase: I changed it to just subtract 2 from the length like `chopped` does.

garakmon: honestly the line I wrote is better because as it was before, it requires the last item in the array have a trailing comma

phase: Okay I changed it to that :ok_hand:

phase:
```c++
src/core/parseutil.cpp:80:53: warning: unknown escape sequence: '\}'
             QList<QString> li = line.remove(QRegExp("[\{|\}]")).remove(" ").split(",");
```
Should this be `\\{`?

garakmon: no but you can just remove the slash altogether

phase: I merged the `remove` calls it and works :ok_hand:

garakmon:
this will fix travis building btw

Issue #18, opened 2018-09-12
Set properties when adding new map


huderlem:
When adding a new map, present the user with a dialog box which allows them to choose things like width, height, tilesets, etc. Currently, it creates new maps with default values.

(originally from https://github.com/yenatch/pretmap/issues/21)

huderlem:
d13a4da50ae3756670bae51db11f5c2bf9213361

Issue #46, opened 2018-10-06
Suggestion: Remember How Porymap's Panels Were Sized Between Sessions


Deokishisu:
Porymap currently forgets how the panels were sized when it is closed, and opens back up to the default sizing. Please have it remember between sessions.

huderlem:
d13a4da50ae3756670bae51db11f5c2bf9213361

PR #54, opened 2018-10-14
Remember window sizes, add new map popup window

garakmon:
~~WIP because there is a (pretty big) bug I need to fix. Maybe somebody can help with it?~~
Will close Issues #18 and #46


src/ui/newmappopup.cpp
```diff
@@ -0,0 +1,103 @@
...
+    layout->blockdata_label = QString("%1_MapBlockdata").arg(newMap->name);
+    layout->blockdata_path = QString("data/layouts/%1/map.bin").arg(newMap->name);
+
+    // EMERGENCY (MUST FIX):causes a segfault...
+    // how do I add new event to newMap without behavior being undefined???
+    if (this->ui->checkBox_NewMap_Flyable->isChecked()) {
+        Event *healSpot = new Event;
+        healSpot = Event::createNewEvent("event_heal_location", newMapName);
+        healSpot->put("map_name", newMapName);
+        newMap->addEvent(healSpot);
```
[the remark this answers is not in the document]
huderlem: Why add a heal location to every new map?  It probably should not any any heal spots.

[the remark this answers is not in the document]
garakmon: there is a checkbox in the popup box for one

garakmon:
this is what it looks like currently
image

src/ui/newmappopup.cpp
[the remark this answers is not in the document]
huderlem: ah, that's more reasonable

huderlem:
This one has been open a long time (my fault).  any chance this can be brought up to date with the new configuration settings?

garakmon:
yep

garakmon:
dang it

garakmon:
phew

garakmon:
okay all up-to-date 👍

src/mainwindow.cpp
```diff
@@ -208,14 +208,13 @@ void MainWindow::loadUserSettings() {
...
-    QSettings settings;
-
-    if (settings.contains("saved_widget_geometry")) {
-        this->restoreGeometry(settings.value("window_geometry").toByteArray());
-        this->restoreState(settings.value("window_state").toByteArray());
-        this->ui->splitter_map->restoreState(settings.value("map_splitter_state").toByteArray());
-        this->ui->splitter_events->restoreState(settings.value("events_splitter_state").toByteArray());
-        this->ui->splitter_main->restoreState(settings.value("main_splitter_state").toByteArray());
+    if (porymapConfig.getRestoreWindowGeometry()) {
+        logInfo("Restoring window geometry from previous session.");
```
huderlem: 2 code improvements I'd like to see here:
1. `porymapConfig.getGeometry()` should return a `QMap`, rather than an array.
2. Call `getGeometry()` once, and use the result when restoring state.

src/config.cpp
```diff
@@ -153,6 +192,16 @@ void PorymapConfig::setPrettyCursors(bool enabled) {
     this->save();
 }
 
+void PorymapConfig::setGeometry(QByteArrayList geometry) {
+    this->restoreGeometry = true;
```
huderlem: This function should either take an argument for each parameter, or there should be aseparate function for each geometry that porymap saves.  Relying on a properly formatted array is more fragile and unreadable.

src/config.cpp
```diff
@@ -130,9 +146,32 @@ QMap<QString, QString> PorymapConfig::getKeyValueMap() {
...
+    map.insert("restore_geometry", this->restoreGeometry ? "1" : "0");
+    map.insert("window_geometry", unicodeByteArray(this->windowGeometry));
+    map.insert("window_state", unicodeByteArray(this->windowState));
+    map.insert("map_splitter_state", unicodeByteArray(this->mapSplitterState));
+    map.insert("events_splitter_state", unicodeByteArray(this->eventsSlpitterState));
+    map.insert("main_splitter_state", unicodeByteArray(this->mainSplitterState));
     return map;
 }
 
+QString PorymapConfig::unicodeByteArray(QByteArray bytearray) {
```
huderlem: Call this `stringFromByteArray()`

src/config.cpp
```diff
@@ -130,9 +146,32 @@ QMap<QString, QString> PorymapConfig::getKeyValueMap() {
     map.insert("recent_map", this->recentMap);
     map.insert("pretty_cursors", this->prettyCursors ? "1" : "0");
     map.insert("map_sort_order", mapSortOrderMap.value(this->mapSortOrder));
+    map.insert("restore_geometry", this->restoreGeometry ? "1" : "0");
```
huderlem: I don't think having this `restore_geometry` boolean is necessary to save.  I don't think there is a case where you'd wan to disable it. (and there is no place in the UI to disable it)

src/config.cpp
huderlem: You'd then have to check to make sure the geometries had non-null/zero values before setting window state on program start, rather than relying on this boolean.

src/mainwindow.cpp
```diff
@@ -731,19 +743,56 @@ void MainWindow::onOpenMapListContextMenu(const QPoint &point)
...
+}
+
+void MainWindow::onNewMapCreated() {
+    QString newMapName = this->newmapprompt->map->name;
+    int newMapGroup = this->newmapprompt->group;
+    Map *newMap_ = this->newmapprompt->map;
+
+    Map *newMap = editor->project->addNewMapToGroup(newMapName, newMapGroup, newMap_);
+
+    qDebug() << "Created a new map named" << newMapName;
```
huderlem: use logInfo() instead of qDebug

src/mainwindow.cpp
```diff
@@ -731,19 +743,56 @@ void MainWindow::onOpenMapListContextMenu(const QPoint &point)
...
+    QStandardItem *newMapItem = createMapItem(newMapName, newMapGroup, numMapsInGroup);
     groupItem->appendRow(newMapItem);
     mapListIndexes.insert(newMapName, newMapItem->index());
 
-    setMap(newMapName);
+    setMap(newMapName, true);
+
+    if (newMap->isFlyable == "TRUE") {
+        addNewEvent("event_heal_location");
+        editor->project->saveHealLocationStruct(newMap);
```
huderlem: Ideally we'd just save one time in this function, but this is doing an extra save just for the heal location.  an saveAllDataStructures() be move down here, or was there a complication with that?

src/config.cpp
garakmon: It seems that anything invalid will just restore to the default settings. I've put several random junk things in the config for these and it never messed up.

src/ui/newmappopup.cpp
```diff
@@ -0,0 +1,98 @@
...
+        this->ui->lineEdit_NewMap_Name->setPalette(QPalette());
+    }
+}
+
+void NewMapPopup::on_pushButton_NewMap_Accept_clicked() {
+    Map *newMap = new Map;
+    MapLayout *layout = new MapLayout;
+
+    // If map name is not unique, use default value. Also replace spaces with underscores.
+    QString newMapName = this->ui->lineEdit_NewMap_Name->text().replace(" ","_");
```
huderlem: This regex would be better: `[a-zA-Z0-9_]+`

src/mainwindow.cpp
garakmon: Well it was a few months ago but I'm 70% sure there was a reason for this -- I will try to determine.

garakmon: Oh yes, it was because I couldn't save the event with existing functions (`addNewEvent`) while the map was not set in the editor, and I had to save before setting the map. I suppose I could try to change this if you wnat?

huderlem: it's fine for now.  was just wondering if it could easily be removed here.

src/mainwindow.cpp
```diff
@@ -736,21 +736,48 @@ void MainWindow::onOpenMapListContextMenu(const QPoint &point)
...
+        QMenu* menu = new QMenu(this);
+        QActionGroup* actions = new QActionGroup(menu);
+        actions->addAction(menu->addAction("Add New Map to Area"))->setData(secName);
+        connect(actions, SIGNAL(triggered(QAction*)), this, SLOT(onAddNewMapToAreaClick(QAction*)));
+        menu->exec(QCursor::pos());
+    } else if (itemType == "map_layout") {
+        QString layoutName = selectedItem->data(Qt::UserRole).toString();
+        QMenu* menu = new QMenu(this);
+        QActionGroup* actions = new QActionGroup(menu);
+        actions->addAction(menu->addAction("Add New Map with Layout"))->setData(layoutName);
```
huderlem: I think they should all just say "Add New Map", since you specify all of the properties in the new pop up window.

src/ui/newmappopup.cpp
```diff
@@ -15,18 +16,40 @@ NewMapPopup::NewMapPopup(QWidget *parent, Project *project) :
...
+            useLayout(layout);
+            setDefaultValues(group, QString());
+            break;
+    }
+}
+
+void NewMapPopup::useLayout(QString mapName) {
+    this->changeLayout = true;
+    this->layoutName = mapName;
+    qDebug() << "will use layout of map" << layoutName;
```
garakmon: remove

PR #59, opened 2018-11-16
[WIP] YAML-style config

garakmon:


aaaaaa123456789:
What's the point? YAML seems _extremely_ overkill for what a simple key/value file would do well...

```
key1=foo
key2=42
key3=bar baz quux
```

garakmon:
I mean, that's basically what I've done, hence "yaml-style".  I guess it could be called anything

PR #77, opened 2019-01-05
Region Map Editor

garakmon:
Adds functionality to: edit the region map background image tilemap, edit the region map layout, edit and create the city maps (the zoomed in ones on the pokenav), change the dimensions of the region map, with other convenience features.

Things that are out of the scope of this PR: adding new map sections, assigning city maps to overworld maps (code is not decompiled yet).


huderlem:
1. This should be its own window, similar to Tileset Editor.
2. What are the two bottom panels for?

garakmon:
1. Okay, can do. (but there is an effect when you open a map from the left list and the layout is open)
2. 
image

3. I'll fix the commits too lol

garakmon:
but it looks terrible because they are so tiny

huderlem:
You can scale those up a lot.  That's how the tileset editor works.

garakmon:
Is it important that saving an unmodified project in porymap results in an OK ROM? I ask because currently I am storing some key:values in a map (I could change to a list of QPairs or something if needed?) and if they are saved, they will not preserve their original order.

aaaaaa123456789:
Can't you just save their order in an auxiliary array/map?

garakmon:
yes

huderlem:
Yes, it should result in an OK ROM, so the order should be preserved.

huderlem:
The region map editor window should warn about unsaved changed, just like the tileset editor.

huderlem:
Feedback that should be addressed:
- Edit-Undo doesn't work properly when click-dragging to paint the region map background image.
- Can't click-drag to paint the city map
- The Region Map Editor window can't be resized
- The "save" button in the City Map context should not exist.  All saving should be via ctrl+S or File > Save
- If I open and save without modifying anything, it writes some bad characters to file:
```
-static const u8 sMapName_PokemonMansion[] = _("POKéMON MANSION");
+static const u8 sMapName_PokemonMansion[] = _("POKÃ©MON MANSION");
```
- The "Set Values" button should be removed, and saving should just be handled with ctrl+S or File > Save
- The City Map dropdown doesn't have any values in it, and it isn't' populated either.
- Click-dragging will let you select out-of-bounds squares in the Map Layout
-

garakmon:
for the record:
requires pret/pokeemerald@af2c398 and pret/pokeruby#721

garakmon:
Okay I believe I have fixed every bug I know about and all of the features seem to be working as intended so I guess this is ready for a real review.

src/ui/regionmapeditor.cpp
```diff
@@ -0,0 +1,660 @@
...
+    this->ui->graphicsView_Region_Map_BkgImg->setFixedSize(this->region_map->imgSize().width() * pow(scaleUpFactor, scaleRegionMapImage - 1) + 2,
+                                                           this->region_map->imgSize().height() * pow(scaleUpFactor, scaleRegionMapImage - 1) + 2);
+    this->ui->graphicsView_Region_Map_Layout->setFixedSize(this->region_map->imgSize().width() * pow(scaleUpFactor, scaleRegionMapImage - 1) + 2,
+                                                           this->region_map->imgSize().height() * pow(scaleUpFactor, scaleRegionMapImage - 1) + 2);
+
+    if (zoom_in) {
+        this->ui->graphicsView_Region_Map_BkgImg->scale(scaleUpFactor, scaleUpFactor);
+        this->ui->graphicsView_Region_Map_Layout->scale(scaleUpFactor, scaleUpFactor);
+    } else {
+        //
```
garakmon: delet

include/project.h
```diff
@@ -47,6 +47,8 @@ class Project
     Map* loadMap(QString);
     Map* getMap(QString);
 
+    QMap<QString, QString> *mapSecToMapHoverName;
```
garakmon: nullptr

src/ui/regionmapeditor.cpp
```diff
@@ -0,0 +1,660 @@
...
+#include "config.h"
+
+#include <QDir>
+#include <QDialog>
+#include <QDialogButtonBox>
+#include <QFormLayout>
+#include <QLineEdit>
+#include <QColor>
+#include <QMessageBox>
+#include <QDialogButtonBox>
```
garakmon: delete

src/core/regionmap.cpp
```diff
@@ -0,0 +1,316 @@
...
+    bfile.write(data);
+    bfile.close();
+}
+
+void RegionMap::saveOptions(int id, QString sec, QString name, int x, int y) {
+    resetSquare(id);
+    int index = getMapSquareIndex(x + this->padLeft, y + this->padTop);
+    if (!sec.isEmpty()) {
+        this->map_squares[index].has_map = sec == "MAPSEC_NONE" ? false : true;
+        this->map_squares[index].secid = static_cast<uint8_t>(project->regionMapSections->indexOf(sec));
```
garakmon: this requires map sections to be defined without any gaps in values, and without duplicates. do both of these requirements make sense? investigate

src/core/regionmap.cpp
```diff
@@ -0,0 +1,316 @@
...
+                mapSecToMapEntry.insert(sec, entry);
+            }
+        }
+        this->map_squares[index].x = x;
+        this->map_squares[index].y = y;
+        this->map_squares[index].duplicated = false;
+    }
+}
+
+void RegionMap::resetSquare(int index) {
```
garakmon: add `resetAllSquares()`?
or `resetLayout()` and `resetImage()`? (clear?)

src/ui/citymappixmapitem.cpp
```diff
@@ -0,0 +1,104 @@
...
+    int y = static_cast<int>(pos.y()) / 8;
+    int index = getIndexAt(x, y);
+    data[index] = static_cast<uint8_t>(this->tile_selector->selectedTile);
+
+    draw();
+}
+
+void CityMapPixmapItem::mousePressEvent(QGraphicsSceneMouseEvent *event) {
+    QPointF pos = event->pos();
+    int x = static_cast<int>(pos.x()) / 8;
```
garakmon: x and y are unused here, delete

src/ui/regionmappixmapitem.cpp
```diff
@@ -0,0 +1,72 @@
...
+    int y = static_cast<int>(event->pos().y()) / 8;
+    emit this->hoveredRegionMapTileChanged(x, y);
+}
+
+void RegionMapPixmapItem::hoverLeaveEvent(QGraphicsSceneHoverEvent *) {
+    emit this->hoveredRegionMapTileCleared();
+}
+
+void RegionMapPixmapItem::mousePressEvent(QGraphicsSceneMouseEvent *event) {
+    QPointF pos = event->pos();
```
garakmon: pos unused

huderlem:
I think this needs to be updated so that the change takes place whenever the user is editing those fields--not when they hit "Save".  Otherwise, everyone will have the same confusion I just did.

Some more context is in Discord.

huderlem:
More bugs I'm noticing:
1. The Map Layout picture doesn't update when you "undo" changes to the Background Image.
2. Don't paint on the "mouse release" event.  Otherwise, you accidentally paint on the Background Image when selecting from the "Edit" menu
3. "Delete Square" doesn't do anything.
4. I reeally think the the x/y spinners should be removed entirely, or made readonly.  Using them to move a map square is a confusing user experience.
5. The "Tools" menu has nothing in it.

huderlem:
Also, we should either add the ability to import new images for the map and cities, or add a label displaying the images' filepaths, so people know where to find them.

garakmon:
1. fixed
2. I was never able to replicate this (probably related to macOS menu bars being detached) but I think it is fixed
3. fixed
4. removed
5. added tile image imports, removed unused generator code
6. if I understood this correctly, then I fixed it
7. not sure when that broke :/ but I'll hide it for now anyways

huderlem:
I tried importing a random image for region_map tiles, and got this error, which makes no sense:
image

huderlem:
If I import a 16-color image, I see an error logged, but no pop-up error dialog:
The provided image has a palette with 16 colors. You must provide an indexed imaged with a 256 color palette."

huderlem:
Minor thing: Remove the space character before the ellipses in "Import Region Map Image Tiles ..." and city image.

Issue #70, opened 2019-01-02
Region Map Sections: do not support "/* ... */" comments


jmwielandt:
Hi, i've comment some lines on my `include/constants/region_map_sections.h` file with `/* ... */` but porymap still reading them.
![image](https://user-images.githubusercontent.com/37153467/50581495-ff8e8b00-0e38-11e9-80bd-aa818d521a2a.png)
![image](https://user-images.githubusercontent.com/37153467/50581502-0cab7a00-0e39-11e9-9b9a-6c285a4fad6a.png)
![image](https://user-images.githubusercontent.com/37153467/50581534-4bd9cb00-0e39-11e9-871f-deed50ce599f.png)

Thx.

garakmon:
For now, using `//` comments should work, but this is a pretty easy fix.

huderlem:
a2e3d17dbb7a86e5b81778f60020ce2284b71ca0

PR #91, opened 2019-01-16
Small Things

garakmon:
There was an outline over the first metatile on the map when the project was first opened regardless of the settings.
Added fix for #70  and a macOS install guide


garakmon:
jk this doesn't work yet

Issue #55, opened 2018-10-14
Implement some kind of Zoom or scaling for Metatiles.


Gamer2020:
Decided to use this on my surface tablet which has a high resolution. Noticed that the tiles were really small and didn't find a way to zoom in on them.

![image](https://user-images.githubusercontent.com/6243575/46919058-6e311a80-cfa7-11e8-9271-57cc75d6b290.png)

phase:
Ctrl + Scrolling the mouse wheel will zoom in the map tiles.

garakmon:
I think they mean the metatiles on the right

Gamer2020:
Yes on the right. I can zoom in on the map no problem but the metatiles are extremely small.

Deokishisu:
I'd like to add to this issue. Not only should the metatiles in the right panel be zoomable in the map view, but also the panels the tileset editor. Gamer2020 will run into the same problem there.

Gamer2020:
Confirmed. Not as small but still the same thing.

![image](https://user-images.githubusercontent.com/6243575/46920625-1c948a00-cfbf-11e8-8709-f3b955541988.png)

Gamer2020:
Zoomed in picture for reference of size difference.

![image](https://user-images.githubusercontent.com/6243575/46920628-3d5cdf80-cfbf-11e8-8430-b7377760d9e8.png)

Gamer2020:
Sweet, I'll try to test tonight.

Gamer2020:
Just tested. Works great, thank you!

Gamer2020:
Actually, just noticed this. Tolerable but just thought I should point it out.

![image](https://user-images.githubusercontent.com/6243575/52923840-b9fa2480-32f7-11e9-8563-523a5c142aa5.png)

PR #97, opened 2019-02-09
add slider to scale metatiles view

garakmon:
#55


huderlem:
Rather than adding zoom functionality in a ton of different places (i.e next the tileset editor would need this), is there a better solution that will solve it for high resolution displays?  This page has some info: https://doc.qt.io/qt-5/highdpi.html

Issue #100, opened 2019-02-23
BUG: picking tiles, scrollbar do whatever it wants


jmwielandt:
Single click to pick a tile and scrollbar goes up or down, selecting multiple tiles.

![image](https://user-images.githubusercontent.com/37153467/53291564-f6b09a80-3793-11e9-86b1-5cc5b85f1ffe.png)

Gamer2020:
I can confirm this. Haven't had a chance to report it, seems to be something with the zoom feature.

Get Outlook for Android

________________________________

garakmon:
okay I've identified the issue

tustin2121:
Any plans on including this in a bugfix release soon? (Especially for those of us who use porymap on Windows, seeing as there's no instructions on how to build on Windows...?)

huderlem:
@tustin2121 likely this weekend

PR #101, opened 2019-02-26
Bug fixes

garakmon:
- #100 fixed
- extended metatile selector scale to apply to border and current selection
- fixed map icons


garakmon:
I also noticed that in the events tab, there are some fields that could be displayed with a SpinBox or a LineEdit instead of the NoScrollComboBox but that would require more than a couple of lines to be changed and it isn't really *that* big of a deal so I'd like a second opinion on whether that would be a worthwhile change.

Issue #64, opened 2018-12-24
[Suggestion] Creation of New Groups and New Tilesets


Gamer2020:
Hi.

I'm sure this is planned for eventually and I apologize if there is already an open item for it. I'd like to suggest the ability to create new/blank groups and new/blank tilesets similar to how a new map can be created.

Deokishisu:
Creating new groups is a simple change in the data/maps/groups.inc file. You can add or remove groups more quickly than any possible porymap UI doing that. Make sure that after you edit the groups.inc file, you open porymap and Save All immediately so that porymap automatically fixes your include/constants/maps.h file to correspond with your new groupings.

SBird1337:
@Deokishisu hmm that sounds not like a very sane process, because you kind of need a tool to fix your `maps.h` but you also kind of don't use a tool and do it manually.

Also: manually editing things like that, especially when normally used with a GUI is always prone to errors.

Gamer2020:
Agree that new groups isn't as bad. 

However, adding new tiles sets seemed tricky to me and I couldn't get it to work manually but I could have been doing something wrong.

huderlem:
Closing, as the remaining work for this Issue is covered by #181

PR #105, opened 2019-03-22
Tileset workflow

SBird1337:
Add a quick dialog to add a new tileset, create it with a dummy image and dummy metatiles for later on editing.

Although this does not touch the mapgroups, it resolves part of #64.

Feedback appreciated.


huderlem:
A few things that need to be addressed:
1. I was able to create two identically-named tilesets `gTileset_Foo` by creating one as a primary and the other as a secondary.  The duplicate name check needs to extend to all tilesets--not just the primary or secondary sets.
2. Can you add ellipses to "New Map" and "New Tileset" menu options? ("New Map..." and "New Tileset...")
3. In the error message for "tileset already exists", include the name of the tileset.  Something like "Tileset 'gTileset_Foo' already exists".
4. Either show a success message after the tileset has been created, or set the current map's tileset to the newly-created tileset.
5. The New Map window is completely broken.  Did you accidentally change that file?  Seems like it shouldn't have been touched by this PR.

SBird1337:
Those should be fixed, I also added a keyboard shortcut (CTRL+SHIFT+N).

SBird1337:
Changed this to WIP, when compiling the tilesets seem to be broken in-game. Currently I have no idea of why that happens, granted it worked before, too.

SBird1337:
The process works fine now, the serialization just misaligned the padding, see 91be88e.

PR #108, opened 2019-04-04
Add a way to name metatiles, for use in c code

Phlosioneer:
Metatiles can be given a name. The name becomes a constant in
includes/contstants/metatile_labels.h.

The plan is to be able to reference metatiles in code using a macro
like `METATILE(Building, TV_ON, Primary)`, which will evaluate to the
value 0x003, or `METATILE(BrendansMaysHouse, MOVING_BOX_OPEN, Secondary)`,
which will evaluate to the value 0x270.

I'm open to suggestions on this PR, but I think there should definitely be a way to make a constant for a metatile. There are a lot of raw metatile values in C code. This implementation lets us pick and choose which ones need constants, and to do so in a way that's modder-friendly and not fragile.

### DIFF
--- a/forms/tileseteditor.ui
+++ b/forms/tileseteditor.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>700</width>
-    <height>600</height>
+    <height>700</height>
    </rect>
   </property>
   <property name="focusPolicy">
@@ -198,6 +198,16 @@
              <bool>false</bool>
             </property>
             <layout class="QGridLayout" name="gridLayout_3">
+             <item row="4" column="0" colspan="3">
+              <widget class="QLabel" name="label_8">
+               <property name="text">
+                <string>Metatile Label (Optional)</string>
+               </property>
+              </widget>
+             </item>
+             <item row="5" column="0" colspan="3">
+              <widget class="QLineEdit" name="lineEdit_metatileLabel"/>
+             </item>
              <item row="2" column="0" colspan="3">
               <widget class="QLabel" name="label_5">
                <property name="text">
--- a/include/core/metatile.h
+++ b/include/core/metatile.h
@@ -3,6 +3,7 @@
 
 #include "tile.h"
 #include <QImage>
+#include <QString>
 
 class Metatile
 {
@@ -12,6 +13,7 @@ public:
     QList<Tile> *tiles = nullptr;
     uint8_t behavior;
     uint8_t layerType;
+    QString label;
 
     Metatile *copy();
     void copyInPlace(Metatile*);
--- a/include/project.h
+++ b/include/project.h
@@ -78,6 +78,7 @@ public:
     void loadTilesetAssets(Tileset*);
     void loadTilesetTiles(Tileset*, QImage);
     void loadTilesetMetatiles(Tileset*);
+    void loadTilesetMetatileLabels(Tileset*);
 
     void saveLayoutBlockdata(Map*);
     void saveLayoutBorder(Map*);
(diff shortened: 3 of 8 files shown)


garakmon:
I think this is a good idea as well. 
I'll give you some feedback though because I played around with it a bit:
1. I couldn't give any labels to metatiles in the secondary tileset (like the moving box in your example).
2. I don't think '-' or '.' should be allowed characters.
3. I can't seem to remove a label from the tileset editor by clearing the lineEdit.
4. Is it a good idea to delete any label that doesn't have a corresponding metatile? For instance, if I open metatile_labels.h and add `#define METATILE_General_TOO_BIG 0x666`, it will be removed by porymap on save. I can't think of a specific use case for such a define, and honestly I don't have a strong opinion on this, but I thought it was worth mentioning.

Phlosioneer:
It deletes everything it doesn't recognize to be a little more robust to metatile size changes. (e.g. if you define a label for the last metatile and then reduce the number of metatiles in the tileset)

I'll investigate the secondary-tileset bug.

I lifted the character-validation code off of the same code that validates event objects' flag names. The bug is probably present in both places. I'll fix that.

I think I know what's going wrong with clearing the lineEdit. I'll fix that too.

Phlosioneer:
Fixed the empty-line behavior, and removed `.` and `-` from the regex.

I couldn't reproduce any issues with assigning labels to secondary tilesets.

garakmon:
After further testing, it doesn't seem to be an issue exclusive to the secondary tileset. It happens when I haven't clicked off of the label or pressed ENTER on my keyboard to "set" the label, so it doesn't save. So to reproduce, just type a label and hit save without changing focus from the lineEdit. 
Probably has something to do with the `editingFinished()` signal, so maybe it doesn't make sense to change the behavior here.

Phlosioneer:
Hmm... The three alternatives are:
- Check for the lineEdit's value when saving. This isn't a good separation of concerns, as the saving code is not part of the tile editor code.
- Figure out a way to remove focus from all controls when saving. This will force `editingFinished` signals from lineEdits.
- Update the internal label whenever the user presses a character (use `textEdited` instead).

`textEdited` is probably the best solution.

garakmon:
I would have suggested using `textEdited`, but it doesn't make sense to add a commit each time this signal is emitted. So, perhaps move only the label setting to that function.

Phlosioneer:
I thought the save code was separate from the tileeditor, but it actually goes through a function in that class. Now it checks the metatile label field before saving.

huderlem:
This needs to disallow identical labels in the same tileset  I was able to name two metatiles "tv", and only the second was saved to the header

huderlem:
The metatiles in the secondary tileset should be saved as their real values (i.e. starting at 0x200).
```
#define METATILE_General_flower   0x004
#define METATILE_General_tv       0x003
#define METATILE_Petalburg_door1  0x048
#define METATILE_Petalburg_door2  0x049
#define METATILE_Petalburg_shrub  0x002
```

In the above example, the Petalburg metatiles should be 0x248, 0x249, and 0x202.  (Use the `NUM_TILES_IN_PRIMARY` for that value)

huderlem:
Everything else seems good-to-go.

Phlosioneer:
I stored the relative values for the tiles because I was concerned that a tileset could be used as a Primary in one map, and a Secondary in another.

garakmon:
Tilesets are either primary or secondary.

Phlosioneer:
Alright, if we have that guarantee, I'll use it 👍

huderlem:
This is exposed in porymap's `Tileset` struct's `is_secondary` field.

garakmon:
a note so I don't forget:
I think it makes most sense to keep labels outside of the range of metatiles.
For example, `gTileset_SecretBase` is a primary tileset, but I am going to label the metatiles in all of the secondary secret base tilesets as being a part of the primary tileset. Examining the code and tilesets will show you why.

garakmon:
actually maybe it is not worth labeling the secret base metatiles

garakmon:
okay this should be good to go now

Issue #117, opened 2019-04-12
Region Map Editor: Part 2


huderlem:
The data specified by `gRegionMapEntries` should also be editable in porymap.  In a new tab, the user should be able to cycle through each of the region map sections.  For the current region map section, the map would show one thing:
1. A rectangular region that can be moved and resized to represent how large the region map section is.

Then, the top-left corner is the x/y coords of `gRegionMapEntries`, along with the rectangle's width/height.

huderlem:
It might be possible to merge this functionality into the existing map layout tab, but I'm not sure how to make that work well.

huderlem:
Another approach (probably the right approach) is to automatically determine these properties from the existing map layout tab.  A good enough algorithm might be:
```
for each region_map_sec_id:
    find the top-left-most occurrence
    find the bottom-right-most occurrence
    x/y = top-left
    width/height = bottom-right - top-left
```
If the user is writing non-rectangular regions for any map sec id, then it wouldn't work in-game anyway

PR #119, opened 2019-04-13
Add gRegionMapEntries Editing / Viewing Tab

garakmon:
addresses #117


huderlem:
This seems to be working great.  Only request I have is that the highlighted map entry should be draggable, to update the x/y coords.  Width/height is fine the way it is.

garakmon:
gotcha

PR #121, opened 2019-04-14
update some read/write functions to use globals instead of statics

garakmon:
This is a breaking change that requires pret/pokeruby#736 and pret/pokeemerald#687.


DizzyEggg:
Why? I'm of the opinion that if it can, it should be static. I am against this PR unless you give a reason as to why get rid of the static variables.

huderlem:
The correct fix is for porymap to respect any symbol qualifiers, and emit them when saving.  In reality, we can't expect everything upstream or in someone's project to be fixed over the long-term.

PR #129, opened 2019-05-05
Don't Crash When Reading Bad Defines

garakmon:
but warn the user (give a file name and line number) about bad expressions.

Reading tilesets now doesn't give warnings about the secret base tileset pointers.

Added colorful log output for Linux and macOS users (and probably this applies to WSL). Windows I can't reliably test and it would require a refactor that is probably not worth it.

### DIFF
--- a/include/core/parseutil.h
+++ b/include/core/parseutil.h
@@ -7,34 +7,12 @@
 #include <QString>
 #include <QList>
 #include <QMap>
+#include <QRegularExpression>
 
 enum TokenType {
     Number,
     Operator,
     Error,
-};
-
-class DebugInfo {
-public:
-    DebugInfo(QString file, QStringList lines) {
-        this->file = file;
-        this->lines = lines;
-    };
-    QString file;
-    int     line;
-    bool    err;
-    QStringList lines;
-    void error(QString expression, QString token) {
-        int lineNo = 0;
-        for (QString line_ : lines) {
-            lineNo++;
-            if (line_.contains(expression)) {
-                this->line = lineNo;
-                break;
-            }
-        }
-        logError(QString("%1:%2: unknown identifier found in expression: '%3'.").arg(file).arg(line).arg(token));
-    }
 };
 
 class Token {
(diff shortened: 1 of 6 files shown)


src/project.cpp
```diff
@@ -1563,94 +1547,89 @@ void Project::readInitialFacingDirections() {
...
-        // The defines should to be sorted by their underlying value, not alphabetically.
-        // Reverse the map and read out the resulting keys in order.
-        QMultiMap<int, QString> definesInverse;
-        for (QString defineName : defines.keys()) {
-            definesInverse.insert(defines[defineName], defineName);
-        }
-        *definesToSet = definesInverse.values();
-    } else {
-        logError(QString("Failed to read C defines file: '%1'").arg(filepath));
+void Project::readCDefinesSorted(QString filename, QStringList prefixes, QStringList* definesToSet) {
```
huderlem: I feel like the `readCDefines` functions should live in `parseUtil.cpp`.  Agree?

include/core/parseutil.h
```diff
@@ -33,10 +60,11 @@ class Token {
 class ParseUtil
 {
 public:
-    ParseUtil();
+    ParseUtil(QString, QString);
     void strip_comment(QString*);
     QList<QStringList>* parseAsm(QString);
     int evaluateDefine(QString, QMap<QString, int>*);
+    DebugInfo *debug;
```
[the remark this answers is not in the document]
garakmon: gah. when I first wrote it I thought it was going to be shared

src/project.cpp
garakmon: yes

huderlem:
This parser object pattern is not good.  I guess `Project` should have a `Parser` object.  Whenever, a new project is opened, `Parser.root` should be set to the new project path.

huderlem:
There are several places where a `filename` variable is created, but only used in the following `parser.whatever(filename)` call.  Can you try to clean those up so it's just simply `parser.whatever("my/file.txt")`.

PR #133, opened 2019-05-21
Map Image Exporter

garakmon:
The window:
image

An example of an image exported (checked: Objects, Grid, Connections:Right):
image

### DIFF
--- a/forms/mapimageexporter.ui
+++ b/forms/mapimageexporter.ui
@@ -33,7 +33,7 @@
           <rect>
            <x>0</x>
            <y>0</y>
-           <width>337</width>
+           <width>373</width>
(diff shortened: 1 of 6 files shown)


huderlem:
The checkbox sections don't anchor to the top, like they should:
image

huderlem:
If you select a connections checkbox, and then hit "Reset", the border isn't cleared in the preview.

huderlem:
The exported image always has a 1-pixel grey border on the left and top sides.

huderlem:
There is a memory leak when opening the "Export Map Image" window a bunch of times.  Total memory usage seems to increase by ~1MB every time I open and close the window.

PR #135, opened 2019-05-22
Flip Full Selection In Tileset Editor

garakmon:
gif

### DIFF
--- a/src/ui/tileseteditor.cpp
+++ b/src/ui/tileseteditor.cpp
@@ -211,6 +211,16 @@ void TilesetEditor::onSelectedTilesChanged() {
 }
 
 void TilesetEditor::onMetatileLayerTileChanged(int x, int y) {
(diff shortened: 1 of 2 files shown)


huderlem:
I just noticed a regression when painting a greater than 1x1 selection onto the Bottom/Top layers.  If I click on the bottom-right corner tile of the Bottom layer, then it paints like this:
image

It looks like this is also a problem in the `master` branch, but maybe you could try and fix that now if it's easy.

garakmon:
is the intended behavior to just draw the top-left corner of the selection onto that corner?

huderlem:
It's intended to only draw the top row of the tile selection in this case.  So, just the top of the sign.

garakmon:
oh "of the bottom layer"

garakmon:
I'm 99.9% sure it is fixed now, I messed around a bunch with it and it seems to work as intended.

Issue #137, opened 2019-06-02
Heal locations are not being saved properly


huderlem:
It seems to be appending numbers to every heal location constant--probably a regression. 

![image](https://user-images.githubusercontent.com/2229288/58761462-3fea5600-850a-11e9-99ca-04738c26a0d9.png)

garakmon:
14a92c4c16cf16aee6298b439bb3b143ae417dd3

PR #138, opened 2019-06-04
Small Bugs

garakmon:
clear `flyableMaps` before reading heal locations,
fix line break parsing of C defines


garakmon:
I tried actually for the last two days to reproduce #137 but I can't. But hopefully this prevents it from happening. I'll wait for confirmation before closing the issue though

PR #141, opened 2019-06-10
beautify collision overlays

nicholatian:
the collision overlays currently shipping with porymap are somewhat unsightly imo. I put together an alternative layout that looks better, following a weathermap-like gradient where blues are passable and reds are not. this is how it looks in action:
screencap


garakmon:
I have to disagree that these collision tiles are an improvement. The current version has more useful icons for the special collision types (surfable water, bridges, elevation transfer, etc.) that these do not. 
Perhaps, if the demand is there, porymap could eventually support some custom ui elements?

SBird1337:
Just because I came across this: It is not reall easy to see height difference with the current setup and with this one as well I think.

Finding and fixing those kind of errors in your collision map can be very tedious. Probably the height should be visible from the colormap as well.

PR #144, opened 2019-07-03
Wild Encounters Editing

garakmon:
I tried to make this user-friendly and as accommodating as possible to different changes people may want to make. I hope I was successful.

Working on documentation now before I forget all of the stuff I added.


garakmon:
whoops
TODO: add config option to disable for those who don't want to use JSON for encounters

huderlem:
forgot about this. will review tonight

SBird1337:
Some things I noticed:
image

* long names are not ideal (See Screenshot)
* UX for adding new groups that do not exist yet is not really intuitive (Maybe add a tab anyways and create a button there, or at least add a note/tooltip that say "Right Clickt to add" or something

SBird1337:
Newly created tabs do not correctly show "0" for the encounter rate:
image
Also the label should probably say "%"

PR #152, opened 2019-08-14
Themes

garakmon:
A little qol enhancement :)

To change themes, View>Themes...>Apply

dark theme:
image


huderlem:
I'm not comfortable with the current state of these themes.  Some problems I'm noticing:
1.  The checkmarks in any checkbox are not visible in all of the non-default themes
2. The font color in combobox dropdowns is always too close to the background color.
3. Something is wrong with the margins of the toolbar dropdowns. The same behavior happens with the comboboxes used in the `up/down/left/right` combobox in the Connections tab.
image

4. `midnight` and `dark` should be the only non-default themes, as the others are pretty obnoxious..

PR #166, opened 2019-09-29
Fixes for Wild Encounter Editing

### LINKED ISSUE #160: Update all encounters and UI when modifying number of slots
Currently, nothing happens visually when the user changes the number of slots in a wild encounter type.  If the user shrinks the number of fields, the UI should update to reflect that and all wild mon data for that type should be shrunk.  Same thing for when the user increases the number of fields.

### LINKED ISSUE #161: Minimum slot ratio value should be 1
The minimum value for the "Slot Ratio" should be 1, not 0.

### LINKED ISSUE #164: Wild Encounter Levels are Changed to Level 0
This is the first time I'm using the 'Issues' option on GitHub so I apologise in advance if I'm not doing this correctly, but the wild encounter levels of any vanilla map you change using the program get changed to level 0.

garakmon:
Outstanding: 
- ~~updating pokeemerald and pokeruby `src/data/wild_encounters.json`, `src/data/wild_encounters.json.txt`, and `src/wild_encounters.c`~~
- ~~update the docs images~~

Adds Project Files section to docs
Fixes Windows build error
Closes #161 
Closes #160 
Closes #164


huderlem:
This seems functionally read-to-go.  However, I want a warning text to appear if the slot chance total is greater than 255 (one byte).  This is because the engine's logic picks a Random number that is a `u8` value.
```
u8 rand = Random() % ENCOUNTER_CHANCE_LAND_MONS_TOTAL;
```
Please add a warning message here:
image

PR #170, opened 2019-11-02
Event Tab Update

garakmon:
Merges the Map and Event tabs into a stacked widget and adds functionality to the pencil, move, and shift toolbuttons to edit events. One potentially not obvious advantage of this is that the event object view and metatiles view share the same scene, so will not be out of sync when changing between tabs.

\* still need to add undo / redo history before this is merged, but opening for feedback.


garakmon:
#195

PR #188, opened 2020-03-06
Specify Json Order When Saving

garakmon:
Order of keys when saving json is now the same order as they were inserted when constructing the object.

Didn't touch most of the parsing code, because the order doesn't matter when it's being read and it's faster, but I can change so that all json io is done with this new OrderedJson if that is preferred.

### DIFF
--- a/include/core/orderedjson.h
+++ b/include/core/orderedjson.h
@@ -56,10 +56,15 @@
 
 #include <QString>
 #include <QVector>
+#include <QFile>
+#include <QTextStream>
 
 #include <map>
 #include <memory>
 #include <initializer_list>
+
+// temp
+#include <iostream>
 
 #ifdef _MSC_VER
     #if _MSC_VER <= 1800 // VS 2013
@@ -158,10 +163,14 @@ public:
     const Json & operator[](const QString &key) const;
 
     // Serialize.
-    void dump(QString &out) const;
-    QString dump() const {
+    void dump(QString &out, int *) const;
+    QString dump(int *indent = nullptr) const {
         QString out;
-        dump(out);
+        if (!indent) {
+            int temp = 0;
+            indent = &temp;
+        }
+        dump(out, indent);
         return out;
     }
 
@@ -199,6 +208,24 @@ private:
     std::shared_ptr<JsonValue> m_ptr;
 };
 
+class JsonDoc {
+public:
+    JsonDoc(Json *object) {
+        this->m_obj = object;
+        this->m_indent = 0;
+    };
+
+    void dump(QFile *file) {
+        QTextStream fileStream(file);
+        fileStream << m_obj->dump(&m_indent);
+        fileStream << "\n"; // pad file with newline
+    }
+
+private:
+    Json *m_obj;
+    int m_indent;
+};
+
 // Internal class hierarchy - JsonValue objects are not exposed to users of this API.
 class JsonValue {
 protected:
@@ -208,7 +235,7 @@ protected:
     virtual Json::Type type() const = 0;
     virtual bool equals(const JsonValue * other) const = 0;
     virtual bool less(const JsonValue * other) const = 0;
-    virtual void dump(QString &out) const = 0;
+    virtual void dump(QString &out, int *indent) const = 0;
     virtual double number_value() const;
     virtual int int_value() const;
     virtual bool bool_value() const;
--- a/include/core/orderedjson.h
+++ b/include/core/orderedjson.h
@@ -55,7 +55,8 @@
 #pragma once
 
 #include <QString>
-#include <vector>
+#include <QVector>
+
 #include <map>
 #include <memory>
 #include <initializer_list>
@@ -88,7 +89,7 @@ public:
     };
 
     // Array and object typedefs
-    typedef std::vector<Json> array;
+    typedef QVector<Json> array;
     typedef std::map<QString, Json> object;
 
     // Constructors for the various types of JSON value.
@@ -152,7 +153,7 @@ public:
     const object &object_items() const;
 
     // Return a reference to arr[i] if this is an array, Json() otherwise.
-    const Json & operator[](unsigned i) const;
+    const Json & operator[](int i) const;
     // Return a reference to obj[key] if this is an object, Json() otherwise.
     const Json & operator[](const QString &key) const;
 
@@ -213,7 +214,7 @@ protected:
     virtual bool bool_value() const;
     virtual const QString &string_value() const;
     virtual const Json::array &array_items() const;
-    virtual const Json &operator[](unsigned i) const;
+    virtual const Json &operator[](int i) const;
     virtual const Json::object &object_items() const;
     virtual const Json &operator[](const QString &key) const;
     virtual ~JsonValue() {}
--- a/include/core/orderedjson.h
+++ b/include/core/orderedjson.h
@@ -1,0 +1,223 @@
+/// Wrappings around QJsonObjects that preserves order
+/* poryjson
+ *
+ * poryjson is a modified version of json11, which adds support to preserve the key order
+ * when dumping json objects, and support to Qt classes
+ *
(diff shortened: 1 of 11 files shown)


src/core/event.cpp
```diff
@@ -239,7 +239,7 @@ void Event::readCustomValues(QJsonObject values)
     }
 }
 
-void Event::addCustomValuesTo(QJsonObject *obj)
+void Event::addCustomValuesTo(OrderedJson::object *obj)
 {
     for (QString key : this->customValues.keys()) {
```
garakmon: TODO: test custom attributes

include/core/orderedjson.h
```diff
@@ -0,0 +1,316 @@
...
+            #define noexcept throw()
+        #endif
+
+        #ifndef snprintf
+            #define snprintf _snprintf_s
+        #endif
+    #endif
+#endif
+
+namespace poryjson {
```
huderlem: Why is the namespace called `poryjson` and not `orderedjson`, or whatever it was initially called?  There doesn't seem to be a reason to give it a strange custom name like that unless you've actually modified the original library.

src/core/event.cpp
```diff
@@ -267,9 +267,9 @@ QJsonObject Event::buildObjectEventJSON()
     return eventObj;
 }
 
-QJsonObject Event::buildWarpEventJSON(QMap<QString, QString> *mapNamesToMapConstants)
+OrderedJson::object Event::buildWarpEventJSON(QMap<QString, QString> *mapNamesToMapConstants)
 {
-    QJsonObject warpObj;
+    OrderedJson::object warpObj;
     warpObj["x"] = this->getU16("x");
```
huderlem: Does `OrderedJson` behave the same way as Qt's library when accessing an element that is non-existent?  (i.e. it doesn't crash, and returns the default value)

src/project.cpp
```diff
@@ -595,78 +604,79 @@ void Project::saveWildMonData() {
...
+            OrderedJson::array subGroupIndices;
             std::sort(fieldInfo.groups[groupName].begin(), fieldInfo.groups[groupName].end());
             for (int slotIndex : fieldInfo.groups[groupName]) {
-                subGroupIndices.append(slotIndex);
+                subGroupIndices.push_back(slotIndex);
             }
             groupsObject[groupName] = subGroupIndices;
         }
-        if (!groupsObject.isEmpty()) fieldObject["groups"] = groupsObject;
+        if (!(groupsObject.isEmpty())) fieldObject["groups"] = groupsObject;
```
huderlem: unnecessary parens were added here

src/project.cpp
```diff
@@ -1501,7 +1512,13 @@ bool Project::readWildMonData() {
     for (auto subObjectRef : wildMonObj["wild_encounter_groups"].toArray()) {
         QJsonObject subObject = subObjectRef.toObject();
         if (!subObject["for_maps"].toBool()) {
-            extraEncounterGroups.insert(subObject["label"].toString(), subObject);
+            QString err;
+            QString subObjson = QJsonDocument(subObject).toJson();
+            OrderedJson::object orderedSubObject = OrderedJson::parse(subObjson, err).object_items();
```
huderlem: Does this need better error handling?  What `OrderedJson::parse(subObjson, err)` fails?

include/core/orderedjson.h
garakmon: I did modify a library I found (`json11`) to save with indentation and not on a single line and use a custom `OrderedMap`. Finding an ordered map library was difficult so I wrote my own that overrides the iterator for `QMap`, but if this solution (somewhat fragile maybe, but imo stable enough for everything we use it for here) isn't acceptable then I could import this one instead https://github.com/Tessil/ordered-map

src/project.cpp
garakmon: It fails with an invalid object, but I don't expect that would ever happen because I don't think `QJsonDocument::toJson` would return something invalid. I added this check here just in case because it was failing in my tests when I had a Json::string without surrounding it with quotes.

src/core/event.cpp
garakmon: Yes

include/core/orderedjson.h
huderlem: Ah, yes.  For some reason, I assumed this was purely an external library.

include/core/orderedjson.h
```diff
@@ -0,0 +1,316 @@
...
+        }
+        return end();
+    }
+
+    bool operator < (const OrderedMap<T>& other) const {
+        return m_keyOrder.size() < other.m_keyOrder.size();
+    }
+
+    // only works for inserting unique keys!
+    T & operator [] (const QString &key) {
```
huderlem: I don't understand what functionality this method is overriding.  Why is it inserting a default value into the underlying `QMap`, rather than an actual value?

huderlem: It appears to be this function? https://doc.qt.io/qt-5/qmap.html#operator-5b-5d

If that is true, then this isn't implemented correctly, since it's supposed to insert only if there is no associated item with `key`.

include/core/orderedjson.h
```diff
@@ -0,0 +1,316 @@
...
+        m_keyOrder.push_back(key);
+        return *QMap<QString, T>::insert(key, T());
+    }
+
+    const T & operator [] (const QString &key) const {
+        return (*find(key)).second;
```
huderlem: This also smells wrong.  There's no need to iterate, as you should just be able to invoke `*QMap<QString, T>::value(key)`, which will have better performance.

include/core/orderedjson.h
```diff
@@ -0,0 +1,316 @@
...
+        for (auto i : *this) {
+            if (i.first == key) {
+                return iterator(this, pos);
+            }
+        }
+        return end();
+    }
+
+    bool operator < (const OrderedMap<T>& other) const {
+        return m_keyOrder.size() < other.m_keyOrder.size();
```
huderlem: Why override this method?  I would thing the existing functionality is sufficient.

include/core/orderedjson.h
huderlem: I'm further confused because this is the only place `m_keyOrder` is appended to, but I would expect you to have overridden `QMap`'s `insert` method, too, and updated `m_keyOrder` there.

include/core/orderedjson.h
garakmon: This operator isn't overloaded in the base class, but it was used in the json library so I had to add something here. I will change the functionality to mimic `std::map` (lexically on the keys afaict) though, I just wasn't sure what that was at the time.

include/core/orderedjson.h
garakmon: Well, it's this way because for porymap we don't try to insert keys that aren't unique or reassign values, and they are only ever inserted with `operator []`. I wanted to keep this as small as possible, maybe not the best idea? When I started this, for some reason I thought Qt already had an ordered map class and was planning to use that. 
I can update this so that it follows a more standard behavior.
Another option is to just use another MIT licensed library that already has all these problems solved. (Or I was thinking I could modify QMap by copying the code from Qt's source, but I do not fully understand the license permissions although I'm pretty sure that's allowed).

huderlem: I want to avoid a situation where unexpected things happen, and this `QMap`-override implementation has some holes that could potentially cause headaches down the road (the risk is admittedly small).  I'd prefer to use https://github.com/Tessil/ordered-map for the `OrderedMap` implementation--it's just a drop-in header, so I wouldn't expect it to be too much work to switch over.

garakmon: I agree this is the safer option. Would you rather have these things in include/core/ or like an include/external for organization purposes?

huderlem:
This is fine once that last `TODO: test custom attributes` comment is addressed.

src/core/event.cpp
garakmon: This works as it should

PR #192, opened 2020-03-11
Add pokefirered support

GriffinRichards:
Makes necessary changes listed below to support using porymap on a pokefirered project.

- Support pokefirered version selection

- Support custom border sizes. Available in config, only enabled by default for pokefirered. Border now extends to fill player view and uses undo/redo

- Allow uniquely named heal locations that differ from the map name

- Support editing respawn data for pokefirered heal locations

- Support pokefirered's other unique event fields (`in_connection` for object events, and `quantity` and `underfoot` for hidden items).

- Support pokefirered floor numbers in map header

- Use new names for header properties (`allow_cycling` and `allow_escaping`)

- Use initial constants instead of hard-coded values for header properties of new maps

- Support pokefirered metatile attributes, including editing encounter type and terrain type

- Use constants for trainer type, allow editing trainer type value directly

- Disable Weather Triggers and Secret Bases for pokefirered

- Disable (for now) the region map editor for pokefirered

- Fix failure to open if there are empty map layouts

- Fix failure to locate tileset folders with underscores in the name

- Prefer dimensions listed in oam tables over base oam for rendering sprites

- Use default tilesets for maps with invalid tilesets. Fixes segfault when creating map using a layout with an invalid tileset

NOTE: To maintain compatibility with pokeemerald/pokeruby, this requires the following PRs:
- pret/pokeemerald#1010
- pret/pokeruby#776

### DIFF
--- a/include/core/map.h
+++ b/include/core/map.h
@@ -17,6 +17,10 @@
 #define DEFAULT_BORDER_WIDTH 2
 #define DEFAULT_BORDER_HEIGHT 2
 
+// Number of metatiles to draw out from edge of map. Could allow modification of this in the future.
+// porymap will reflect changes to it, but the value is hard-coded in the projects at the moment
+#define BORDER_DISTANCE 6
+
 class Map : public QObject
 {
     Q_OBJECT
@@ -65,7 +69,8 @@ public:
     int getBorderHeight();
     QPixmap render(bool ignoreCache, MapLayout * fromLayout = nullptr);
     QPixmap renderCollision(qreal opacity, bool ignoreCache);
-    bool blockChanged(int, Blockdata*);
+    bool mapBlockChanged(int i, Blockdata * cache);
+    bool borderBlockChanged(int i, Blockdata * cache);
     void cacheBlockdata();
     void cacheCollision();
     Block *getBlock(int x, int y);
@@ -82,12 +87,14 @@ public:
     void addEvent(Event*);
     QPixmap renderConnection(MapConnection, MapLayout *);
     QPixmap renderBorder();
-    void setDimensions(int newWidth, int newHeight, bool setNewBlockData = true);
+    void setDimensions(int newWidth, int newHeight, bool setNewBlockdata = true);
+    void setBorderDimensions(int newWidth, int newHeight, bool setNewBlockdata = true);
     void cacheBorder();
     bool hasUnsavedChanges();
 
 private:
     void setNewDimensionsBlockdata(int newWidth, int newHeight);
+    void setNewBorderDimensionsBlockdata(int newWidth, int newHeight);
 
 signals:
     void mapChanged(Map *map);
--- a/include/core/map.h
+++ b/include/core/map.h
@@ -35,6 +35,7 @@ public:
     QString allowRunning;
     QString allowBiking;
     QString allowEscapeRope;
+    int floorNumber;
     QString battle_scene;
     QString sharedEventsMap = "";
     QString sharedScriptsMap = "";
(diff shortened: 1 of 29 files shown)


src/ui/newmappopup.cpp
```diff
@@ -123,7 +135,11 @@ void NewMapPopup::on_pushButton_NewMap_Accept_clicked() {
     newMap->name = newMapName;
     newMap->type = this->ui->comboBox_NewMap_Type->currentText();
     newMap->location = this->ui->comboBox_NewMap_Location->currentText();
-    newMap->song = "MUS_DAN02";
+    if (projectConfig.getBaseGameVersion() == BaseGameVersion::pokefirered) {
+        newMap->song = "MUS_MASARA";
```
garakmon: these should probably pick from the project's list of {songs, weather, etc}

GriffinRichards: Sure. The first elements of both lists in the projects are `WEATHER_NONE` and `MUS_DUMMY`, which are probably better defaults anyway.

GriffinRichards:
Re: Does saving with this version of porymap break pokefirered, no <sup>TM</sup> . Almost all basic functions work properly now. The following are the only remaining major issues I know of:
- Saving a map with a non-standard border size writes the border incorrectly. It is also rendered incorrectly in porymap. (see e.g. Viridian Forest)
- Heal locations as fly destinations are not set up on pokefirered's end. porymap is also not set up to handle pokefirered's actual respawn locations
- Metatile attributes are not parsed correctly. The metatile behaviors in the tileset editor are not displayed correctly as a result

garakmon:
Going to keep a list of issues I encounter here 

- segfault/crash when adding new map with layout UnusedLayout1
    -- due to having invalid tilesets ("NULL"), porymap should handle this

include/core/map.h
```diff
@@ -14,6 +14,13 @@
 #include <QGraphicsPixmapItem>
 #include <math.h>
 
+#define DEFAULT_BORDER_WIDTH 2
+#define DEFAULT_BORDER_HEIGHT 2
+
+// Number of metatiles to draw out from edge of map. Could allow modification of this in the future.
+// porymap will reflect changes to it, but the value is hard-coded in the projects at the moment
+#define BORDER_DISTANCE 7
```
huderlem, line 22: Why 7, and not 6?  6 is what emerald/ruby use.

GriffinRichards: Drawing 7 shows the maximum extent of the player view. If you only draw 6 the player view rectangle extends beyond the edge of what's drawn.

huderlem:
Before merging, please revisit the case where the `use_custom_border_size=0` in a pokefirered project. It might cause bad things to happen in its current implementation.  Either force it to `1` in pokefirered projects or gracefully handle whatever bad thing happens due to non-constant border dimensions.

huderlem:
Looks like `project.cpp` conflicts now that the JSON ordering PR was merged

GriffinRichards:
For `use_custom_border_size` comment see response in Discord. Also found an issue with SS Anne interior maps not loading metatile behaviors because the layout constant `SSANNE_*` is inconsistent with the tileset folder name `ss_anne`. Can fix on pokefirered's end and then add a special case for `SS_ANNE` in `layoutConstantFromName`, same as `SS_TIDAL`.

huderlem:
I guess this is good to go now?  The plan is to update SS Anne's folder in pokefirered, correct?

GriffinRichards:
Good now, porymap was always guessing the path instead of getting it from the header (different spot than RSE)

PR #195, opened 2020-03-14
Merges The Event and Map Views

garakmon:
When switching between the Map and Events tabs, the view stays the same so your position on the screen also stays the same. Also copies the map-shift behavior to shift events, and the pencil is another shortcut to add events.
IMHO this provides a better map-editing experience. 
I've had this done for quite a while, but have been attempting to extend porymap's undo history to all types of edits. It hasn't been going great so I'll just separate those tasks.

@GriffinRichards this will likely conflict with #192 but I can resolve those if this is merged before that


huderlem:
Issues I'm noticing:
1. There is weird extra padding around the main tab view. (Map/Events/Header/etc.)
2. The selected edit mode does not persist when switching between Map and Events.  If I have "Move" mode selected in the Map view, I would expect to still have "Move" mode selected when I go to the Events tab.  Switching between the tabs should persist the mode when it's compatible, and fallback to either Pencil or Pointer, depending on the tab.
3. Edit Connection -> Open Events tab -> Observe the connection is not rendered to reflect the updated connection.  It _does_ update properly when opening the Map tab.

garakmon:
rebased and addressed comments

PR #204, opened 2020-04-18
Add ability to export map stitch images.

huderlem:
The border code in the map image exporter will need to be updated when firered is merged.

### DIFF
--- a/src/ui/mapimageexporter.cpp
+++ b/src/ui/mapimageexporter.cpp
@@ -66,7 +66,7 @@ struct StitchedMap {
 };
 
 QPixmap MapImageExporter::getStitchedImage(QProgressDialog *progress, bool includeBorder) {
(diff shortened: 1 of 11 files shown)


src/ui/mapimageexporter.cpp
```diff
@@ -28,32 +32,192 @@ MapImageExporter::~MapImageExporter() {
...
 }
 
+struct StitchedMap {
+    int x;
+    int y;
+    Map* map;
+};
+
+QPixmap MapImageExporter::getStitchedImage(QProgressDialog *progress, bool includeBorder) {
+    // Do a bread-first search to gather a collection of
```
GriffinRichards, line 67: Fresh out of the oven

huderlem:
The border logic will need to be updated in this PR after the pokefirered works is in master.

PR #226, opened 2020-05-16
Read MAX_MAP_DATA_SIZE

GriffinRichards:
Use `MAX_MAP_DATA_SIZE` in the repo for the map dimensions limit. Does not change the new map prompt's bad width/height limit

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2291,9 +2291,8 @@ void MainWindow::on_pushButton_ChangeDimensions_clicked()
     heightSpinBox->setMinimum(1);
     bwidthSpinBox->setMinimum(1);
     bheightSpinBox->setMinimum(1);
-    // See below for explanation of maximum map dimensions
-    widthSpinBox->setMaximum(0x1E7);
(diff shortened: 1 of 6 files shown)


src/mainwindow.cpp
```diff
@@ -2353,14 +2352,7 @@ void MainWindow::on_pushButton_ChangeDimensions_clicked()
 }
 
 bool MainWindow::mapDimensionsValid(int width, int height) {
```
huderlem, line 2354: This `mapDimensionsValid()` function/logic should really be in `project.cpp`

src/project.cpp
```diff
@@ -2097,6 +2099,24 @@ bool Project::readTilesetProperties() {
         logWarn(QString("Value for tileset property 'NUM_PALS_TOTAL' not found. Using default (%1) instead.")
                 .arg(Project::num_pals_total));
     }
+    it = defines.find("MAX_MAP_DATA_SIZE");
```
huderlem, line 2102: This should probably be its own function.  It doesn't make sense to put it in `readTilesetProperties()`

PR #227, opened 2020-05-17
fix map dimension checks in new map window

garakmon:


huderlem:
Can we use the same error message that's present in the "Change Dimensions" dialog?  Also, the warning frame's size is not aligned with the form:
image

Issue #228, opened 2020-05-19
Suggestion: Add ruler tool


Sierraffinity:
This just makes it easier for me to count metatiles. Not too important.

huderlem:
Does the grid overlay satisfy this need?  The cursor rectangle is slightly awkward when interacting with events because it partially covers them up, so you can't see what's underneath.  It's a minor detail either way--it's hard to know which way is best.

ketsuban:
A ruler tool would also solve @Sierraffinity's problem. Being able to measure distance nondestructively would help when scripting movements.

huderlem:
Good idea, but is that any different than pressing `Ctrl+G` to quickly turn on the grid view?  Maybe it is.

Sierraffinity:
Honestly, a ruler tool would be even better. I didn't even consider that as an option.

Sierraffinity:
Relatedly, could the coordinates be shown in the bottom left corner when mousing over the map, like the draw view?

huderlem:
#310

PR #237, opened 2020-06-02
Show metatile label names in map editor

GriffinRichards:
Mousing over metatiles in the map view will now show the metatile label name in the bottom left, as is already done in the metatile selector


GriffinRichards:
Also tacking on showing the mouse coords in the events view as mentioned in #228

PR #239, opened 2020-06-06
Fix potential crash using editor->map after clearing map cache

GriffinRichards:


src/project.cpp
```diff
@@ -143,6 +143,7 @@ void Project::clearMapCache() {
         Map *map = mapCache->take(mapName);
         if (map) delete map;
```
huderlem, line 144: Why not set `map` to `nullptr` right here, instead?  Seems pointless to hookup the slot action.

huderlem: Oh, this is not the `editor` class

PR #246, opened 2020-06-25
Implement Proper Triple Metatile Layer Support

ultima-soul:


forms/mainwindow.ui
```diff
@@ -233,23 +233,23 @@
         </property>
         <item>
          <layout class="QHBoxLayout" name="horizontalLayout_10">
```
huderlem, line 235: Please revert all changes to this file.

huderlem:
Can we also disable/remove the layer type dropdown menu, since it's pointless with triple layer?

forms/mainwindow.ui
Sierraffinity: I hate how Qt always insists on dicking around with this file when you don't mean to... and also explicitly prevents you from making changes in the text view.

PR #272, opened 2020-07-09
Add duplicate shortcut for events

### LINKED ISSUE #269: Suggestion: Event duplication
It would be useful to have a way to quickly create an identical event to another, perhaps with a keyboard shortcut or click-and-drag-while-holding-down-a-key or even something in the event panel.

GriffinRichards:
Adds Ctrl/Cmd+D as a shortcut to duplicate events. Incidentally this makes it much easier to expose the fact that there's no enforced limit on creating events
Closes #269


GriffinRichards:
Updated to enforce limit imposed by `OBJECT_EVENT_TEMPLATES_COUNT`

PR #275, opened 2020-07-13
Respect x/y flip and palette for external tileset editor tile selections

### LINKED ISSUE #263: Suggestion: Allow for flipping of multiple grabbed tiles even if grabbed from existing metatile
As it stands right now you can flip an entire group of tiles in the tileset editor, but only if you grab them from the tile view. You should be able to do the same with a group grabbed from an existing metatile.

GriffinRichards:
Fixes #263 

The X Flip/Y Flip checkboxes are now always relative to the selected tiles visually, i.e. if you select any tile with the X Flip checkbox on, it will appear in the selection flipped horizontally (regardless of whether it was selected from an existing layer or from the tileset). Tile selections from the layer view can now be manipulated the same as they would for tiles selected from the tileset.

Single tile selections from a layer will no longer highlight the corresponding tile in the tile selector, as the layer tile may be flipped differently than its original tile.

### DIFF
--- a/src/ui/tileseteditortileselector.cpp
+++ b/src/ui/tileseteditortileselector.cpp
@@ -112,16 +112,23 @@ QList<Tile> TilesetEditorTileSelector::buildSelectedTiles(int width, int height,
     QList<QList<Tile>> tileMatrix;
     for (int j = 0; j < height; j++) {
         QList<Tile> row;
+        QList<Tile> layerRow;
         for (int i = 0; i < width; i++) {
             int index = i + j * width;
             Tile tile = selected.at(index);
             tile.xflip ^= this->xFlip;
             tile.yflip ^= this->yFlip;
-            tile.palette = this->paletteId;
             if (this->xFlip)
-                row.prepend(tile);
+                layerRow.prepend(tile);
             else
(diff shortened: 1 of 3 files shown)


src/ui/tileseteditortileselector.cpp
```diff
@@ -95,36 +95,45 @@ void TilesetEditorTileSelector::updateSelectedTiles() {
...
+    QList<Tile> tiles;
+    QList<QList<Tile>> tileMatrix;
+    for (int j = 0; j < height; j++) {
+        QList<Tile> row;
+        for (int i = 0; i < width; i++) {
+            int index = i + j * width;
+            Tile tile = selected.at(index);
+            tile.xflip ^= this->xFlip;
+            tile.yflip ^= this->yFlip;
+            tile.palette = this->paletteId;
```
huderlem, line 120: This is introducing a bug where the palette id of the first-selected tile is applied to every externally-selected tile.

huderlem:
We can be smarter about the horizontal flipping, too.  It should only apply to each individual layer.  This makes it easy to horizontally flip complete metatiles, rather than having to duplicate the layers individually.

Issue #87, opened 2019-01-11
Allow deleting of events with the DEL key, do not allow deletion of undeletable items


SBird1337:
When having an event item selected (object, warp, ...) allow to delete it using the DEL key, additionally hide any delete functionality for items that should not be deleted, e.g. heal spots

garakmon:
closing with #280 merged

Issue #14, opened 2018-09-12
undo/redo buttons


huderlem:
Undo/Repo is already already implemented for painting map tiles. It would be nice to extend this feature to the other editing features of a map.

(originally from https://github.com/yenatch/pretmap/issues/32)

garakmon:
closing with #280 merged

PR #280, opened 2020-07-29
Overhaul Edit History

garakmon:
The intention here is to be able to add edit history for other kinds of edits besides map metatiles (namely, event editing). Qt has a framework that manages this pretty well.

- [x] Removes the current map undo/redo history implementation in favor of the Qt Undo Framework. 
- [x] Add edit commands for event editing (create, delete, move for now)
- [x] Fix the script API


garakmon:
closes #87 
closes #14

PR #287, opened 2020-08-23
Straight paths

BigBahss:
This feature allows painting straight paths by snapping to either the X or Y axis, depending on the first direction the user draws in.

Works for both map tiles and collision tiles and works in conjunction with smart paths. Also works for map shifting. Currently bound to Ctrl to quickly enable it while editing<del>, and also has a check box next to smart paths</del>. Using Ctrl gives you the ability to enable it on-the-fly in the middle of painting. This gives flexibility to tile painting; for instance the user could start painting a normal path, then hold Ctrl to continue that as a straight path, then release Ctrl to continue as a normal path.

### DIFF
--- a/docsrc/manual/editing-map-tiles.rst
+++ b/docsrc/manual/editing-map-tiles.rst
@@ -136,6 +136,11 @@ Smart Paths provide an easy way to paint pathways, ponds, and mountains.  If the
 
     Smart Paths from Right-Click Selection
 
+Straight Paths
+--------------
+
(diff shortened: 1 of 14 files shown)


BigBahss:
After using this feature for a little while I'm starting to feel like the ui checkbox might be unnecessary as it feels a little cumbersome to use over the keyboard shortcut. If you plan to merge this then let me know if you want the checkbox and if not then I can remove it.

garakmon:
I would tend to agree, binding to a held key is sufficient.

huderlem:
I just played around with this.  Works really well--we should have had this a long time ago.  Can you update `docsrc/`, too?

BigBahss:
Thanks, and yeah I planned to update the docs, I was just waiting for feedback. So I'll get on that.

BigBahss:
Updated `docsrc/`, but couldn't figure out how to create a high quality gif from my screen-cap that didn't take up a lot more space than the gifs already in the docs (I'm incompetent). I feel like the description works well enough, but a gif or 2 would make it look more polished.

PR #293, opened 2020-08-29
Retain external selection on map change (closes #291)

### LINKED ISSUE #291: Selection to another Maps not able
You can't copy some part of Maps in Selection to copy in a another Map. In older version was this able.

unknown
unknown-1

BigBahss:


GriffinRichards:
This reintroduces the issue fixed by 79e384ef76b2c6ff855b473057db7663591720bc

BigBahss:
Ah, you're right, but only reintroduces it for external selections. I'll try and fix that.

BigBahss:
Fixed now, when changing tilesets with an invalid external selection it just selects the first tile.

PR #302, opened 2020-09-13
Should Fix Project Open?

garakmon:
I'm failing to convince myself that the fix is any more complicated than simply not calling `initWindow()` more than once. As porymap is now, the old `Project` object is deleted when a new one is opened, and with it all of the existing signals between `Project`-`MainWindow`, and `Project`-`Editor`. So as long as we are not creating the connections again between `MainWindow` and its `ui`, there should never be more than one signal fired per action.


src/mainwindow.cpp
```diff
@@ -1472,6 +1472,7 @@ void MainWindow::addNewEvent(QString event_type)
     if (editor && editor->project) {
         DraggablePixmapItem *object = editor->addNewEvent(event_type);
         if (object) {
+            updateObjects();
```
BigBahss, line 1475: Yeah this was the best solution I was able to come up with as well. I don't love it because it causes `updateSelectedObjects()` to get get called a bunch more times for every new object added, but this is a solution and I didn't notice any real life performance impact.

BigBahss:
I just realized this regresses the disabling of the ui when no project is open.

huderlem:
There's a  lot of uncertainty in the description and comments in this PR.  What should we do about it?

garakmon:
From what I remember this fixed everything that I tested, but I did not go super deep into tracing memory leaks, etc. when somebody opens a project several times in the same session. I think I tested with a broken version of ruby switching between a working project and a broken project (including different versions) and it seemed to work fine.

BigBahss:
`MainWindow::openRecentProject()` is set to always return true if `porymapConfig.getRecentProject()` returns an empty string, this should be changed to false which will fix the disabling of the ui.

BigBahss:
I tested this for memory leaks, this pr improves memory leaks quite a bit. I set up a test that cycled opening emerald, firered, and a broken ruby project 100 times and Porymap's memory usage afterwards was only 1.3 GB, compared to 2.3 GB on master branch (lol).
Once that one thing is fixed I feel like this is good to merge.

PR #304, opened 2020-09-16
Store & restore sub-editors' geometry & state from config

BigBahss:
So this started off as attempting to fix the size policies for the Tileset Editor which would cause all the widgets to display very cramped and overlapped with my theming on Linux and looked like this:

Screenshot_2020-09-16_03-15-58

..Although I couldn't figure out how to do so in a way that didn't drastically alter the spacing, and I didn't want to needlessly mess with the styling on Windows and Mac. So I decided that the geometry & state of the Tileset Editor should just be stored in the config file so that I only had to adjust the window once, rather than every time I open the Tileset Editor. Once I did that I figured why not store & restore the geometries & states of the Region Map Editor and Palette Editor as well.

In addition I removed the `Qt::WA_DeleteOnClose` attribute from the Tileset Editor and Region Map Editor so that the log wasn't spammed with `"Restoring <window> geometry from previous session."` for re-opening those within the same session (which also speeds up the re-opening of those windows). I also gave the Palette Editor its appropriate `windowTitle`.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2659,7 +2659,10 @@ void MainWindow::on_actionRegion_Map_Editor_triggered() {
             return;
         }
         connect(this->regionMapEditor, &QObject::destroyed, [=](QObject *) { this->regionMapEditor = nullptr; });
-        this->regionMapEditor->setAttribute(Qt::WA_DeleteOnClose);
+        logInfo("Restoring region map editor geometry from previous session.");
+        QMap<QString, QByteArray> geometry = porymapConfig.getRegionMapEditorGeometry();
+        this->regionMapEditor->restoreGeometry(geometry.value("region_map_editor_geometry"));
+        this->regionMapEditor->restoreState(geometry.value("region_map_editor_state"));
     }
 
     if (!this->regionMapEditor->isVisible()) {
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1204,8 +1204,11 @@ void MainWindow::on_actionNew_Tileset_triggered() {
 
 void MainWindow::updateTilesetEditor() {
     if (this->tilesetEditor) {
-        this->tilesetEditor->setMap(this->editor->map);
-        this->tilesetEditor->setTilesets(editor->ui->comboBox_PrimaryTileset->currentText(), editor->ui->comboBox_SecondaryTileset->currentText());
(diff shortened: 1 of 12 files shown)


src/mainwindow.cpp
```diff
@@ -2516,7 +2516,10 @@ void MainWindow::on_actionTileset_Editor_triggered()
         this->tilesetEditor = new TilesetEditor(this->editor->project, this->editor->map, this);
         connect(this->tilesetEditor, SIGNAL(tilesetsSaved(QString, QString)), this, SLOT(onTilesetsSaved(QString, QString)));
         connect(this->tilesetEditor, &QObject::destroyed, [=](QObject *) { this->tilesetEditor = nullptr; });
-        this->tilesetEditor->setAttribute(Qt::WA_DeleteOnClose);
```
huderlem, line 2519: By removing `Qt::WA_DeleteOnClose`, is there now a memory leak?  I thought I remember setting that on purpose.

huderlem:
The rest of the PR looks good.

src/mainwindow.cpp
garakmon: Perhaps calling `closeSupplementaryWindows()` somewhere along the line when porymap is being closed (probably the `MainWindow` destructor) is sufficient.
Assuming there are no issues in using the same `TilesetEditor` object for the entire session.

BigBahss: I remember checking that it calls the destructor because I had the same thought, and it does as far as I remember. But this was also 9 days ago, so I'll re-test this and let you know.

garakmon: I guess since it is created as a child of `MainWindow`, it would already get cleaned up by the destructor.

BigBahss: Okay, I just verified, and all the destructors get called whether those windows are still open or not. But I also noticed a slight bug that allows you to close the `TilesetEditor` while the `PaletteEditor` remains open, effectively allowing you to edit the tileset while the window is 'closed'. Then if you re-open the `TilesetEditor` and close it again, it then will prompt you to save those changes. I'll fix that.

BigBahss:
I fixed that bug I mentioned. I also refactored parts of `TilesetEditor`, mostly its initialization and the order of functions. If you don't like the refactorings then just merge up to 3679fa123db6dabba9ff9552cb88879c6e49a873.

PR #306, opened 2020-09-19
Add Import Map from Advance Map 1.92 feature.

ultima-soul:
Very messy and hacky right now so any style or refactoring suggestions are appreciated.

Have to look into FRLG and RS exported maps later. 

You choose the AMap .map file to import, the New Map dialog pops up with map width, map height, tileset labels filled in which are read from the .map header.


ultima-soul:
FRLG and RS support has been implemented and confirmed respectively. It also fills in border width and height obtained from the header in the New Map dialogue if your project is based on pokefirered or has custom border sizes and imports the border metatile data from the .map if present.

forms/mainwindow.ui
```diff
@@ -3034,6 +3040,7 @@
    <class>AdjustingStackedWidget</class>
    <extends>QStackedWidget</extends>
    <header>adjustingstackedwidget.h</header>
+   <container>1</container>
```
huderlem, line 3043: Any idea what this `<container>1</container>` change is?

src/core/mapparser.cpp
```diff
@@ -0,0 +1,116 @@
...
+    int numMetatiles = mapWidth * mapHeight;
+    int expectedFileSize = 20 + (numBorderTiles * 2) + (numMetatiles * 2);
+    if (in.length() != expectedFileSize) {
+        *error = true;
+        logError(QString(".map file is an unexpected size. Expected %1 bytes, but it has %2 bytes.").arg(expectedFileSize).arg(in.length()));
+        return nullptr;
+    }
+
+    Blockdata *blockdata = new Blockdata();
+    for (int i = mapDataOffset; (i + 1) < in.length(); i += 2) {
```
huderlem, line 77: `i < in.length()` should work just as well, right?

src/core/mapparser.cpp
```diff
@@ -0,0 +1,116 @@
...
+        uint16_t word = static_cast<uint16_t>((in[i] & 0xff) + ((in[i + 1] & 0xff) << 8));
```
huderlem, line 78: Why is the `& 0xFF` needed?  Isn't `in[i]` already a single byte?

src/core/mapparser.cpp
```diff
@@ -0,0 +1,116 @@
...
+        blockdata->addBlock(word);
+    }
+
+    Blockdata *border = nullptr;
+    if (numBorderTiles != 0) {
+        border = new Blockdata();
+        for (int i = 20; (i + 1) < 20 + (numBorderTiles * 2); i += 2) {
```
huderlem, line 85: Same here, I think you can just use `i` instead of `i + 1` for the range comparison.

huderlem, line 85: Can you simply check `i < mapDataOffset`?

src/core/mapparser.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+    MapLayout *mapLayout = new MapLayout();
+    mapLayout->width = QString::number(mapWidth);
+    mapLayout->height = QString::number(mapHeight);
+    mapLayout->border_width = (borderWidth == 0) ?  QString::number(2) :  QString::number(borderWidth);
+    mapLayout->border_height = (borderHeight == 0) ?  QString::number(2) :  QString::number(borderHeight);
+
+    QList<QString> tilesets = project->tilesetLabelsOrdered;
+
+    if (mapPrimaryTilesetNum > tilesets.size())
```
huderlem, line 99: Rather than assuming the tilesets are ordered the same between the two projects, would it make more sense to ask the user for the tileset, instead?

src/mainwindow.cpp
```diff
@@ -2319,6 +2339,54 @@ void MainWindow::on_actionExport_Stitched_Map_Image_triggered() {
...
+        msgBox.setText("Failed to import map from Advance Map 1.92 .map file.");
+        QString message = QString("The .map file could not be processed. View porymap.log for specific errors.");
+        msgBox.setInformativeText(message);
+        msgBox.setDefaultButton(QMessageBox::Ok);
+        msgBox.setIcon(QMessageBox::Icon::Critical);
+        msgBox.exec();
+        return;
+    }
+
+    // TODO: This is crude because it makes a history entry for every newly-imported metatile.
```
huderlem, line 2371: Can all this commented out code be removed?  Doesn't the map importer window add a brand new map to the project?  Why is this code trying to do things with metatile history?

src/core/mapparser.cpp
```diff
@@ -0,0 +1,116 @@
...
+    int mapPrimaryTilesetNum = static_cast<unsigned char>(in.at(8)) |
+                               (static_cast<unsigned char>(in.at(9)) << 8) |
+                               (static_cast<unsigned char>(in.at(10)) << 16) |
+                               (static_cast<unsigned char>(in.at(11)) << 24);
+    int mapSecondaryTilesetNum = static_cast<unsigned char>(in.at(12)) |
+                                 (static_cast<unsigned char>(in.at(13)) << 8) |
+                                 (static_cast<unsigned char>(in.at(14)) << 16) |
+                                 (static_cast<unsigned char>(in.at(15)) << 24);
+
+    /*int maxMetatiles = primaryTileset ? Project::getNumMetatilesPrimary() : Project::getNumMetatilesTotal() - Project::getNumMetatilesPrimary();
```
huderlem, line 51: Remove dead code?

src/ui/newmappopup.cpp
```diff
@@ -164,6 +171,86 @@ void NewMapPopup::setDefaultValues(int groupNum, QString mapSec) {
     }
 }
 
+void NewMapPopup::setDefaultValuesImportMap(MapLayout *mapLayout) {
+    ui->lineEdit_NewMap_Name->setText(project->getNewMapName());
```
huderlem, line 175: Most of this function is copy-pasted defaults from the other function in this class.  Let's move those shared defaults into a single function.

src/project.cpp
```diff
@@ -1923,8 +1923,16 @@ Map* Project::addNewMapToGroup(QString mapName, int groupNum, Map *newMap, bool
     if (!existingLayout) {
         mapLayouts.insert(map->layoutId, map->layout);
         mapLayoutsTable.append(map->layoutId);
-        setNewMapBlockdata(map);
-        setNewMapBorder(map);
+        if (importedMap) {
+            map->layout->lastCommitMapBlocks.blocks = new Blockdata;
```
huderlem, line 1927: Is this history-related code needed?  You should just creating a new map in the same pattern as the existing New Map window, which doesn't need history-related logic.

src/core/mapparser.cpp
huderlem: I'm not saying we should do it one way or the other.  I'm just asking the question.

forms/mainwindow.ui
```diff
@@ -3022,6 +3023,11 @@
     <string>Export Map Stitch Image...</string>
    </property>
   </action>
```
huderlem, line 3025: Maybe we should add another separator?
image

huderlem:
Importing seems to work well.  Nice job.  Still have a some code cleanup to do, though.

forms/mainwindow.ui
ultima-soul: No clue.

src/core/mapparser.cpp
ultima-soul: I was reusing the map.bin parsing code for consistency in readBlockdata so that should be looked at perhaps as well.

src/core/mapparser.cpp
ultima-soul: See previous response.

forms/mainwindow.ui
BigBahss: I've noticed Qt Designer will add that (not sure when or why), I've always removed it manually since I didn't know what it was, and it wasn't present in any of the other porymap `.ui` files that I looked at.

src/core/mapparser.cpp
ultima-soul: Good catch

src/mainwindow.cpp
ultima-soul: Sure. I believe I was using the importing metatile .bvd trigger code as a base which may have had this.

src/core/mapparser.cpp
ultima-soul: Yeah, I want to clean this part up by doing maybe a few more proper checks to see if the input is valid. I'll look into it.

forms/mainwindow.ui
Rangi42: I'd expect New Map and New Tileset in the File menu, even though they bring up option dialogs rather than file choosers.

ultima-soul:
Addressed most review comments and merged in the latest commits to resolve conflicts.

src/core/mapparser.cpp
ultima-soul: Turns out this is required since accessing a `QByteArray` returns a `QByteRef` which actually gives an `int`.

PR #310, opened 2020-10-02
Ruler tool

BigBahss:
Adds a ruler tool in the `Events` tab with the `Pointer` tool selected, in order to measure tiles non-destructively for movement scripting, as outlined in #228 . Activated by `right-click`, deactivated by `right-click` again. The ruler's position can locked in place by `left-click` while it is active. The ruler's dimensions are displayed as a `tool tip` and in the `status bar`.

Peek 2020-10-02 15-55

Things that would be nice to fix/add in the future:
- The dimensions `QToolTip` will disappear after a short while, even when specifying a longer `msecDisplayTime`. I'd like for there to be a way to keep it on screen, especially when locking the ruler. But I added the dimensions to the `statusBar` for this reason.
- The ruler will currently draw itself in a fixed way (the x-axis first and then the y-axis). It would be nice if it checked the movement permissions and selected a permissible path if the default is not. This is mostly cosmetic, but would make implementing the next bullet point a little better.
- Have porymap generate the script step instructions in order to copy them to the clipboard. I imagine this could work as simply as pressing `ctrl+c` while the ruler is locked.

I'll might look into those things in the future, but wanted to submit what I have now.

Also, this PR also implements a static function `Metatile::coordFromPixmapCoord()` because the following code was repeated in many places:
```C++
QPointF pos = event->pos();
int x = static_cast<int>(pos.x()) / 16;
int y = static_cast<int>(pos.y()) / 16;
```


src/editor.cpp
```diff
@@ -1967,8 +1967,8 @@ void Editor::objectsView_onMousePress(QMouseEvent *event) {
     if (map_item && map_item->paintingMode != MapPixmapItem::PaintMode::EventObjects) {
         return;
     }
-    if (this->map_edit_mode == "paint" && event->buttons() & Qt::RightButton) {
-        this->map_edit_mode = "select";
+    if (this->obj_edit_mode == "paint" && event->buttons() & Qt::RightButton) {
+        this->obj_edit_mode = "select";
```
BigBahss, line 1962: I think this function originally checking and setting `map_edit_mode` when editing objects may have been a mistake, which would cause the default tool to be the `Pointer` when switching back to the `Map` tab. If this was intentional then let me know.

src/ui/draggablepixmapitem.cpp
```diff
@@ -58,10 +58,21 @@ void DraggablePixmapItem::bindToUserData(QComboBox *combo, QString key) {
...
 void DraggablePixmapItem::mousePressEvent(QGraphicsSceneMouseEvent *mouse) {
     active = true;
-    last_x = static_cast<int>(mouse->pos().x() + this->pos().x()) / 16;
-    last_y = static_cast<int>(mouse->pos().y() + this->pos().y()) / 16;
+    QPoint pos = Metatile::coordFromPixmapCoord(mouse->scenePos());
+    last_x = pos.x();
+    last_y = pos.y();
     this->editor->selectMapEvent(this, mouse->modifiers() & Qt::ControlModifier);
     this->editor->selectingEvent = true;
+    if (this->editor->obj_edit_mode == "select") {
```
huderlem, line 66: all of the logic inside this `if` block really belongs in the `MapRuler` class.

src/ui/mapruler.cpp
```diff
@@ -0,0 +1,137 @@
...
+}
+
+bool MapRuler::isMousePressed(QGraphicsSceneMouseEvent *event) const {
+    return event->buttons() & acceptedMouseButtons() && event->type() == QEvent::GraphicsSceneMousePress;
+}
+
+void MapRuler::showDimensions(const QPoint &screenPos) {
+    // This is a hack to make the tool tip follow the cursor since it won't change position if the text is the same.
+    QToolTip::showText(screenPos + QPoint(16, -8), statusMessage + ' ');
+    QToolTip::showText(screenPos + QPoint(16, -8), statusMessage);
```
huderlem, line 83: It would be nice if the tooltip stayed visible for more than one second.  `QToolTip::showText` takes a "duration" parameter, but I couldn't get it to work.

huderlem:
Looks pretty good to me.  Let's just de-spaghetti that one piece of logic in `draggablepixmapitem.cpp`.

src/ui/draggablepixmapitem.cpp
BigBahss: True, I thought about doing that initially, but I was trying to match the way other `if` blocks in this function did things by parsing inputs in this function. I'll change this.

BigBahss: Wait, my bad, I thought this was the `if` block in `editor.cpp`. In this case I was just matching that `if` block.

huderlem: I'm not sure which one you're talking about.  All I know is the block I commented on should probably be moved into `MapRuler`

src/ui/mapruler.cpp
BigBahss: Yeah, I played around with that and couldn't get it to work either. Earlier in this branch I was rendering the dimensions as a part of the ruler graphic using `QPainter::drawText()` but the problem with that approach was that the size of the text was fixed and was hard to see when zoomed out, and was hard to see against dark map tiles. For those reasons it just seemed cleaner and easier to use a tooltip. I'll play around with this more and try to find a better solution.

src/ui/draggablepixmapitem.cpp
BigBahss: I agree. And the block I was talking about is in `Editor::mouseEvent_map()` which almost exactly matches this `if` block. So if I change this one I'll probably change that one as well.

PR #316, opened 2020-10-14
Focus map zooming on the cursor

### LINKED ISSUE #256: Suggestion: Keep focus on the currently selected tile when zooming in the map view
I tend to use ctrl+scroll to zoom when working on a map. It would be nice if the currently selected tile, i.e. the one the mouse cursor is currently hovering over, was kept in focus when zooming. This is a fairly common feature in many image editors, and it seems like it would be a good addition to porymap.

BigBahss:
Closes #256.
Zooming the map will now zoom into the location of the cursor rather than the top left corner. If the cursor is not on the map then the map will zoom into the center.


src/editor.cpp
```diff
@@ -1481,6 +1506,31 @@ void Editor::createConnectionItem(MapConnection* connection, bool hide) {
     connection_edit_items.append(connection_edit_item);
 }
 
+// Hides connected map tiles that cannot be seen from the current map (beyond BORDER_DISTANCE).
+void Editor::maskNonVisibleConnectionTiles() {
```
garakmon, line 1510: can you explain this please

BigBahss: Yeah, so to enable this functionality I had to expand the `graphicsView` to fill the entire frame, which caused outer parts of connected maps to be displayed (tiles extending further than `BORDER_DISTANCE` from the current map). These outer tiles were previously hidden by resizing the `graphicsView`. This resize was reseting the scroll bars and preventing the zoom from anchoring on the mouse cursor. So to keep the displayed tiles the same I draw over them with the background color of the `graphicsView`, this is the 'mask'.

huderlem:
On initial load, there is a funny gray area around the map:
image

When I switch tabs, the gray area disappears.  Any idea why?  Otherwise seems to work well enough.  Certainly better than before!

BigBahss:
This one perplexes me, I can't recreate the gray area (and have no idea why it would disappear when switching tabs). Does the gray area come back when switching back to the map tab, or does it stay gone? And do you still get an off-colored area when using other porymap themes?

garakmon:
It is drawing the widget while the window is disabled so it is using the widget's disabled color palette. The window is disabled when porymap is loading the project in case it encounters an error. In the MainWindow constructor you could probably add `setWindowDisabled(!isProjectOpen());` and remove `setWindowDisabled(false);` from `initWindow()`.

PR #320, opened 2020-10-26
Fix MapRuler dimensions tooltip

BigBahss:
Polishes the displaying of `MapRuler`'s dimensions by using a `QLabel` as a floating tool-window, displayed in the top-left corner of `graphicsView_Map`, and stays visible until the ruler is deactivated. Also adds the start and end coordinates to the text.

event-tool-ruler


huderlem:
Hmm, I'm not sure this is actually that much better than the current functionality.  Can we not position that windows near the cursor, so the user doesn't have to look away?

BigBahss:
We could have it follow the cursor like originally intended. That's was what I was doing at first, but I realized the stationary window was less obtrusive, since the "following" window could end up obscuring part of the map the user might be interested in (it would require more complicated logic to decide where to position the window based on the direction the mouse is moving in). I felt that the locking mechanism made it so that looking away wasn't inconvenient (position the ruler, lock it where you want it, and check the values). As the ruler currently works the tooltip disappears almost immediately after locking the ruler, meaning the user has to check the status bar for the values, which is less obvious and very easy to miss. I agree that having the values right by the cursor is also convenient, but those were the reasons I made it stationary. I could still change it though, let me know what you think.

BigBahss:
Another possible solution would be to forgo more complicated positioning logic and just allow the window to be dragged out of the way by the user after the ruler is locked. I honestly don't know the best solution here and I'm open to changing it. The stationary window just felt cleaner to me at the time.

PR #323, opened 2020-11-13
Shortcuts Editor

BigBahss:
Adds support for user-customizable shortcuts.

- Adds a new `Shortcut` class that interfaces the same as `QShortcut` but allows multiple key-combinations to trigger the shortcut (like how `QAction`'s allow multiple shortcuts to be set for it).
- The `ShortcutsEditor` and `ShortcutsConfig` are easily extensible. Any newly added `QAction`'s or `Shortcut`'s in the future will automatically be seen by `ShortcutsEditor` and added to `ShortcutsConfig` (no hard-coding config keys required).
- The `ShortcutsEditor` will automatically group shortcuts by their parent window and will duplicate-check within these context groups.
- For a `QAction` or `Shortcut` to be displayed and customizable in `ShortcutsEditor` it needs to have a non-empty `objectName()` and a non-empty `text` or `whatsThis` property. The `objectName()` is used for the config key, and the `text` or `whatsThis` property is displayed next to the shortcuts in the `ShortcutsEditor`.

Tested on Windows and Linux. Should be tested on Mac due to the different keyboard layout.

Peek 2020-11-13 01-16


BigBahss:
Shortcut editing has been added to the docs

Issue #189, opened 2020-03-08
Allow deleting wild encounters for a map


huderlem:
Currently, you can only add wild encounters to a map that doesn't have any encounters.  You can't do the opposite, so a user has to manually edit `wild_encounters.json` to remove encounters for a map.

PR #324, opened 2020-11-14
Implement delete for wild encounter table

V-FEXrt:
Fixes https://github.com/huderlem/porymap/issues/189

### DIFF
--- a/src/editor.cpp
+++ b/src/editor.cpp
@@ -376,13 +376,13 @@ void Editor::deleteWildMonGroup() {
     if (msgBox.clickedButton() == deleteButton) {
         auto it = project->wildMonData.find(map->constantName);
         if (it == project->wildMonData.end()) {
-          logError(QString("Failed to find data for map %s. Unable to delete").arg(map->constantName));
+          logError(QString("Failed to find data for map %1. Unable to delete").arg(map->constantName));
           return;
         }
 
         int i = project->encounterGroupLabels.indexOf(labelCombo->currentText());
         if (i < 0) {
-          logError(QString("Failed to find selected wild mon group: %s. Unable to delete")
+          logError(QString("Failed to find selected wild mon group: %1. Unable to delete")
                    .arg(labelCombo->currentText()));
           return;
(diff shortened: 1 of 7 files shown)


V-FEXrt:
Discord: VFEXrt#1871

src/editor.cpp
```diff
@@ -357,6 +357,35 @@ void Editor::addNewWildMonGroup(QWidget *window) {
...
+    QPushButton *deleteButton = msgBox.addButton("Delete", QMessageBox::DestructiveRole);
+    msgBox.addButton(QMessageBox::Cancel);
+    msgBox.setDefaultButton(QMessageBox::Cancel);
+    msgBox.exec();
+
+    if (msgBox.clickedButton() == deleteButton) {
+        auto it = project->wildMonData.find(map->constantName);
+        it.value().erase(labelCombo->currentText());
+
+        int i = project->encounterGroupLabels.indexOf(labelCombo->currentText());
```
huderlem, line 380: Do we need to be more defensive for this logic?  For example, is it possible for `it` to be null, or for `i` to be `-1`?  I just want to be sure this can't crash.

V-FEXrt: Yeah you are right. Both have out of bounds values that I should've checked. Updated.

src/editor.cpp
```diff
@@ -375,13 +375,22 @@ void Editor::deleteWildMonGroup() {
 
     if (msgBox.clickedButton() == deleteButton) {
         auto it = project->wildMonData.find(map->constantName);
-        it.value().erase(labelCombo->currentText());
+        if (it == project->wildMonData.end()) {
+          logError(QString("Failed to find data for map %s. Unable to delete").arg(map->constantName));
```
huderlem, line 379: I don't think that `%s` is a valid format argument and would cause a crash.  To my knowledge, you can only use the numeric args, like `%1` or `%2`.  https://doc.qt.io/qt-5/qstring.html#arg

Same thing for the other QString below.

PR #325, opened 2020-12-01
Open map scripts directly to the script of the selected event

BigBahss:
- ~~The `Open Map Scripts` button~~ can now open the map's scripts file directly to the script of the ~~first~~ selected event if the user provides the shell command for their preferred text editor. The user provides the command with `%F` and `%L` that will be substituted with the file name and line number. Porymap parses the scripts file to find the line number of the script definition (works for .inc and poryscript, including `raw` tags).
- Adds a `PreferenceEditor` window to more easily add new preferences in the future, and moves themes to this window.
- Adds a menu action `Tools -> Open Project in Text Editor` for convenience, and adds a corresponding command to preferences where `%D` will be substituted with the project root directory. The menu action will be disabled if the user hasn't provided this command.

Screenshot_2020-12-01_18-06-27


BigBahss:
The original behavior of the `Open Map Scripts` button has been restored in favor of adding a dedicated tool-button next to the `Script` combo-box. The tool-button will only appear if the user has set the `Goto Line Command`.

go-to-script-button

BigBahss:
The tool-button now works for any event script. I updated docs, changelog, and UI to reflect that.

src/editor.cpp
```diff
@@ -2024,6 +2026,67 @@ void Editor::deleteEvent(Event *event) {
...
+}
+
+bool Editor::startDetachedProcess(const QString &command, const QString &workingDirectory, qint64 *pid) const {
+    logInfo("Executing command: " + command);
+#ifdef Q_OS_WIN
+    // On Windows, a QProcess command must be wrapped in a cmd.exe command.
+    const QString program = QProcessEnvironment::systemEnvironment().value("COMSPEC");
+    QStringList arguments({ QString("/c"), command });
+#else
+    QStringList arguments = QProcess::splitCommand(command);
```
huderlem, line 2084: I'm having a lot of trouble getting this to work on Windows.  For example, this is failing:
```
"C:\Program Files\Sublime Text 3\sublime_text.exe" %F:%L
```
```
Executing command: "C:\Program Files\Sublime Text 3\sublime_text.exe" D:/cygwin64/home/huder/pokeemerald/data/maps/PetalburgCity/scripts.inc:103
'\"C:\Program Files\Sublime Text 3\sublime_text.exe\"' is not recognized as an internal or external command,
operable program or batch file.
```

Modifying the `startDetached()` call to simply `return QProcess::startDetached(command);` seems to make things simpler.  Is there any reason you opted for the 4-param version of `startDetached()`?

Also, `QProcess::splitCommand()` is a Qt 5.15 addition.

BigBahss: `QProcess::startDetached(command)` has been deprecated which is why I avoided it and used the 4 param version, I could've left out the latter 2 params since they have default values, but I added them for the sake of code re-use. I can also remove `QProcess::splitCommand()`. I'll look into why the above example is failing.

BigBahss:
Finally convinced Windows to cooperate. The commands work for absolute program paths now. I tested Sublime and VSCode executables both as absolute paths and when they exist in the Path variable.

Also, since splitting the shell command isn't as simple as splitting on spaces, I grabbed the source for `QProcess::splitCommand()` and put it in a static `ParseUtil::splitShellCommand()` function. I added a macro around the function that checks the Qt version, and if its >= 5.15 it just calls `QProcess::splitCommand()` instead.

Issue #267, opened 2020-07-07
Suggestion: Allow manual entry of RGB Values in Tileset Palette Editor


vexio:
Instead of using the sliders (which can be annoying to fine-tune sometimes, or even more annoyingly to swap two color slot values), allow the user to click to the right of the RGB Value slider and manually enter a number for each value. 

![porymap_hvyvc1MGY7](https://user-images.githubusercontent.com/20468242/86691284-89369000-bfbd-11ea-877d-5d9cc4fc8048.png)

Obviously, if the entered value is not divisible by 8 (or is 0), it will round up/down to the nearest valid number.

aaaaaa123456789:
Ideally, instead of a simple "divide by 8" conversion, the program could do the proper conversion (multiply by 31, divide by 255, round to nearest).

> Obviously, if the entered value is not divisible by 8 (or is 0), it will round up/down to the nearest valid number.

0 is both valid and divisible by any integer other than itself, including 8.

vexio:
> 0 is both valid and divisible by any integer other than itself, including 8.

Right right, that was a misplacement/typo. Meant to say 0 would be an exception.

aaaaaa123456789:
> > 0 is both valid and divisible by any integer other than itself, including 8.
> 
> Right right, that was a misplacement/typo. Meant to say 0 would be an exception.

Why?

SBird1337:
`rgb5 = round((rgb * 31) / 255)` works for every integer in the given interval [0;255]

PR #331, opened 2020-12-15
Added hex boxes to set palettes.

marinaluna:
I find that when I try to insert color codes or rearrange colors it is incredibly tedious to play with the sliders. This lets you copy/paste hex codes into a box and set the colors :) It looks like [https://imgur.com/jMFk8aY](this).

### DIFF
--- a/forms/paletteeditor.ui
+++ b/forms/paletteeditor.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>817</width>
-    <height>739</height>
+    <width>909</width>
+    <height>918</height>
    </rect>
   </property>
   <property name="focusPolicy">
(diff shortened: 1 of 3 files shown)


Jaizu:
Would you mind adding an input to set individual values as well? I find those better to work with than using HEX codes, at least from my experience in rom hacking with a few graphical editors (Aseprite, Character Maker Pro)
It would fix #267 as well.

marinaluna:
That shouldn't be hard at all, I'll look into it tomorrow.

huderlem:
We should probably just remove the RGB horizontal sliders in favor of RGB number inputs.

Deokishisu:
Please don't remove the palette sliders. They're great for subtly tinkering with colors without having to dig into each value. Having boxes to input the color is great, but there should be both. Also, A-Map has both.
unknown

garakmon:
merged into #432

PR #343, opened 2021-01-30
Add completion suggestions for global event scripts

BigBahss:
Typing into the `Script` combo-box will now suggest all global event scripts, in addition to the map's scripts. Expanding the combo-box will still only show the map's scripts since showing a huge list of global scripts isn't very convenient.

### DIFF
--- a/include/core/parseutil.h
+++ b/include/core/parseutil.h
@@ -60,6 +60,9 @@ public:
     static int getScriptLineNumber(const QString &filePath, const QString &scriptLabel);
     static int getRawScriptLineNumber(QString text, const QString &scriptLabel);
     static int getPoryScriptLineNumber(QString text, const QString &scriptLabel);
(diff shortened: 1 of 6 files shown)


huderlem:
I'm noticing a small bug where selecting an event object and typing in a script (but not finalizing it?) duplicates or insert it into the available autocompletion options.  As you can see here, there are a bunch of extra option from when I was typing in the "FatMan" script.
image

BigBahss:
Interesting, I'll look into that. (Also I misclicked to close lol).

garakmon:
Any interest in revisiting this? The issue mentioned above seems like it could be fixed by changing the insert policy to NoInsert for the combobox.

BigBahss:
I was putting off revisiting this because I started working on another project, but I'll try what you have suggested. Thanks for looking into that.

BigBahss:
Fixed

PR #345, opened 2021-02-03
Store recent map in ProjectConfig instead of PorymapConfig

BigBahss:
Recent maps are stored per project.


huderlem:
It's unclear why we've had it in the base config this whole time...

BigBahss:
Yeah I was just browsing the code and had that realization.

PR #346, opened 2021-02-14
Upgrade signal/slot connections

BigBahss:
Upgrades all string-based connections to functor-based because they are checked at compile time and give better performance. Besides that, IDE's don't understand the `SIGNAL` and `SLOT` macros, and it's difficult to find references of functions within them (not even Qt Creator understands it, funnily enough).

Reference: https://doc.qt.io/qt-5/signalsandslots-syntaxes.html


huderlem:
Should this PR go to `master` instead of `qt6`?  I'm not sure how soon `qt6` will be ready for primetime.

BigBahss:
It could, and that was my initial plan, but I noticed some of the connections I was changing had other changes in this branch, like changing the `valueChanged()` signals to `textChanged()`. I can just make those changes so we can merge this into `master`.

BigBahss:
Alright, updated.

PR #348, opened 2021-02-14
Refactor Blockdata

BigBahss:
Refactors `Blockdata` to simply inherit `QVector<Block>` which simplifies a lot of code as it only needs the `serialize()` member function. Additionally, all usages of pointers to `Blockdata` have been removed to fix/prevent memory leaks. There shouldn't be any runtime impact from using values over pointers, since Qt's shared-data classes are just pointers that copy-on-write (for this reason, all other usages of pointers to shared-data classes should eventually be changed).

Reference for shared-data classes: https://doc.qt.io/qt-5/implicit-sharing.html

This shouldn't be merged until I or someone else reviews/tests the code further. I intend review this in-depth later this evening.

### DIFF
--- a/src/ui/mappixmapitem.cpp
+++ b/src/ui/mappixmapitem.cpp
@@ -76,7 +76,7 @@ void MapPixmapItem::shift(QGraphicsSceneMouseEvent *event) {
 }
 
 void MapPixmapItem::shift(int xDelta, int yDelta, bool fromScriptCall) {
-    Blockdata oldMetatiles = !fromScriptCall ? map->layout->blockdata : Blockdata();
+    Blockdata oldMetatiles = map->layout->blockdata;
 
     for (int i = 0; i < map->getWidth(); i++)
     for (int j = 0; j < map->getHeight(); j++) {
--- a/src/ui/mappixmapitem.cpp
+++ b/src/ui/mappixmapitem.cpp
@@ -76,7 +76,7 @@ void MapPixmapItem::shift(QGraphicsSceneMouseEvent *event) {
 }
 
 void MapPixmapItem::shift(int xDelta, int yDelta, bool fromScriptCall) {
-    Blockdata *backupBlockdata = new Blockdata(*map->layout->blockdata);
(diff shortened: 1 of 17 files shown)


BigBahss (commented):
Alright I reviewed everything and marked the files that don't necessarily need further review. I think someone should review the rest of the files because some of them previously involved null-checks and I'm not entirely sure if `Blockdata.isEmpty()` is functionally equivalent to a null-check in those cases.

BigBahss:
The files that should be reviewed are `map.cpp`, `project.cpp`, `bordermetatilespixmapitem.cpp`, `collisionpixmapitem.cpp`, and `mappixmapitem.cpp`.

src/ui/mappixmapitem.cpp
```diff
@@ -89,20 +90,12 @@ void MapPixmapItem::shift(int xDelta, int yDelta, bool fromScriptCall) {
         destY %= map->getHeight();
 
         int blockIndex = j * map->getWidth() + i;
-        Block srcBlock = backupBlockdata->blocks->at(blockIndex);
+        Block srcBlock = oldMetatiles.at(blockIndex);
```
huderlem, line 93: I think the logic is incorrect here.  If `fromScriptCall` is `true`, then this will segfault due to out-of-bounds access.

huderlem: Also, I just noticed that the caller of this function isn't passing `fromScriptCall = true`.  Could you tack that onto this PR?  It's in `mainwindow_scriptapi.cpp`.

huderlem:
Everything else looks good (and seems to be behaving correctly when I build it).

src/ui/mappixmapitem.cpp
BigBahss: Yeah you're right. I assumed this was following the same structure as the other functions that copy the old metatiles, but didn't think about the fact that this one accesses them. And yes I'll make that change in `mainwindow_scriptapi.cpp`.

PR #349, opened 2021-02-15
Convert pointer members in Project

BigBahss:
Part of the effort to fix/prevent memory leaks, this converts all the unnecessary pointers in `Project` to values. This one is much more "find and replace" than #348.

### DIFF
--- a/include/core/parseutil.h
+++ b/include/core/parseutil.h
@@ -44,15 +44,15 @@ public:
     static QString readTextFile(QString);
     static int textFileLineCount(const QString &path);
     void strip_comment(QString*);
-    QList<QStringList>* parseAsm(QString);
+    QList<QStringList> parseAsm(const QString &filename);
     int evaluateDefine(QString, QMap<QString, int>*);
     QStringList readCArray(QString text, QString label);
     QMap<QString, QString> readNamedIndexCArray(QString text, QString label);
     QString readCIncbin(QString text, QString label);
     QMap<QString, int> readCDefines(QString filename, QStringList prefixes, QMap<QString, int> = QMap<QString, int>());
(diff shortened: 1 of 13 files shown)


BigBahss:
I updated this PR to correct some memory leaks in `ParseUtil`.

include/core/parseutil.h
```diff
@@ -39,42 +39,41 @@ class Token {
...
     QString root;
     QString text;
     QString file;
-    QList<Token> tokenizeExpression(QString expression, QMap<QString, int>* knownIdentifiers);
-    QList<Token> generatePostfix(QList<Token> tokens);
-    int evaluatePostfix(QList<Token> postfix);
-    void error(QString message, QString expression);
+    QList<Token> tokenizeExpression(QString expression, const QMap<QString, int> &knownIdentifiers);
+    QList<Token> generatePostfix(const QList<Token> &tokens);
+    int evaluatePostfix(const QList<Token> &postfix);
```
huderlem, line 75: What's the reason you changed so many function signatures to use `const &`?

huderlem: The rest looks good to me.  I'm just not sure why a lot of these are using `const &` now.  When the object is copied, it probably just the reference container, and not the underlying map or list items themselves.

BigBahss: It's true that copying `Qt`'s shared-data classes is less expensive than copying the corresponding `std` class for instance, but it's generally a good idea not to copy if you don't need to. There is *some* overhead to copying `Qt`'s data classes, even if it is minimal. I think maybe the best argument for passing by `const &` in these cases is that all of `Qt`'s own functions do so for container classes. My general rule of thumb is to only pass containers by value if the function actually needs a copy.

PR #350, opened 2021-02-16
Use QPointer when we need to set a nullptr on destruction

BigBahss:
`QPointer`'s automatically set themselves to nullptr when the `QObject` is destroyed, so I changed the sub-windows to use them since we were making a connection to achieve the same thing.


BigBahss:
This one's pretty simple, I'm just going to merge it.

PR #351, opened 2021-02-16
Metatile leaks

BigBahss:
`Metatile` was leaking memory because it had no destructor to free it's `QList<Tile> *`. ...But it doesn't need one if we just get rid of the pointer. Also, I changed it to just use an implicitly defined copy constructor and copy assingment operator.


huderlem:
This one looks good to me, but I want to do some testing on it before merging.

BigBahss:
Yeah that's a good idea. Also I noticed that `Tileset` has the exact same problem and I've been working on sorting that out and I was going to tack that on to this.

huderlem:
merge conflict

PR #359, opened 2021-03-07
Song Option in New Map Popup

hjk321:
The new map popup lets you configure most header data from the get-go when making a map. However, the option to select the song from this window is absent. This fixes that.


BigBahss:
Tested it, looks good. I'll wait in case someone else wants to review it.

hjk321:
...and as a special preorder bonus, poryscript scripts now generate with a newline at the end of the file!

PR #368, opened 2021-05-25
Autosave

BigBahss:
Adds an auto-save feature which is configured in the `PreferenceEditor`. Auto-save can be set up to save a map after a set delay, when changing maps, or both. Auto-save doesn't create any temporary files, it saves directly to the project files. Due to this, there is a prominent warning in the preferences not to enable auto-save if the user is not using git.

Screenshot_2021-05-25_11-02-51

### DIFF
--- a/forms/preferenceeditor.ui
+++ b/forms/preferenceeditor.ui
@@ -84,6 +84,23 @@
          </property>
         </widget>
        </item>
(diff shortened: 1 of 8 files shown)


huderlem:
Closing this for now.  Will maybe revisit autosave in the future.

PR #376, opened 2021-09-08
A Few More Issues

### LINKED ISSUE #305: editing graphics is not detected as a change
Repro:
1. Open a map.
2. Go to the events tab.
3. Select an object.
4. Change the object's sprite.

At this point, observe that there is no `*` in the window's title. Likewise, doing File -> Save does not make any changes to any files. In order to get the change saved, you must edit some other part of the map (i.e. by moving the sprite).

I'm on Windows, but I suspect this is not a Windows specific issue.

### LINKED ISSUE #337: Changing music in map header doesn't mark a map as dirty
Title. Reproed on the preview build sent to me yesterday while testing #336.

### LINKED ISSUE #347: Editing connections should be undo-able or at least mark the map as dirty
Similar to #305

### LINKED ISSUE #362: Bug: Map combo boxes should be sanitized
Currently, it's possible to enter hidden characters like newlines into the various comboboxes in events (and probably map headers, etc.).  This can allow copy-paste errors to make their way into the final assembly files, which results in invalid assembler syntax.

For example, the following results in the assembler to output a literal newline in the event's macro, which is invalid syntax:
image

garakmon:
Validate combo boxes to disallow junk characters (I couldn't think of an instance where we would need to allow special characters to be typed into a combobox, but if they exist we should remove the validator)

closes #362 

Mark map as changed for remaining kinds of edits to maps (only known exception is an event custom attribute edit).

closes #305
closes #337
closes #347

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -256,7 +256,7 @@ void MainWindow::initEditor() {
     connect(this->editor, &Editor::currentMetatilesSelectionChanged, this, &MainWindow::currentMetatilesSelectionChanged);
     connect(this->editor, &Editor::wildMonDataChanged, this, &MainWindow::onWildMonDataChanged);
     connect(this->editor, &Editor::mapRulerStatusChanged, this, &MainWindow::onMapRulerStatusChanged);
-    connect(this->editor, &Editor::shouldUpdateWindow, this, [=]() { showWindowTitle(); });
+    connect(this->editor, &Editor::editedMapData, this, &MainWindow::markMapEdited);
     connect(ui->toolButton_Open_Scripts, &QToolButton::pressed, this->editor, &Editor::openMapScripts);
     connect(ui->actionOpen_Project_in_Text_Editor, &QAction::triggered, this->editor, &Editor::openProjectInTextEditor);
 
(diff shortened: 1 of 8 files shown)


hjk321:
Tested, both features confirmed working on qt6 for mac.

src/ui/noscrollcombobox.cpp
```diff
@@ -16,6 +16,10 @@ NoScrollComboBox::NoScrollComboBox(QWidget *parent)
     setEditable(true);// can set to false manually when using
     this->completer()->setCompletionMode(QCompleter::PopupCompletion);
     this->completer()->setFilterMode(Qt::MatchContains);
+
+    QRegularExpression re("[A-Za-z_0-9]+");
```
huderlem, line 20: I'm worried this regex is too restrictive.  Perhaps it would be better to just exclude whitespace?  I'm not sure.

src/mainwindow.cpp
```diff
@@ -782,34 +797,44 @@ void MainWindow::on_comboBox_Song_currentTextChanged(const QString &song)
 {
     if (editor && editor->map) {
         editor->map->song = song;
+        editor->map->hasUnsavedDataChanges = true;
```
huderlem, line 800: It might be nice to move these two coupled lines into a shared function or similar.

src/ui/noscrollcombobox.cpp
garakmon: it seems like these are all labels, constants, or numbers, but it is probably better to be cautious here I could me missing something

garakmon: although you'd still be able to type '\n' into the box, you can't paste any spaces

huderlem:
Looks good to me now.

PR #387, opened 2021-11-20
Log more script errors

GriffinRichards:
Log script errors in more situations. This handles 3 notable cases:
1. If an error occurs in a function called by `setTimeout`.
2. If Porymap is unable to find the function specified by the name given to `registerAction` (e.g. if it was not exported).
3. If `addImage` fails to load the image at the given filepath.

The manual was also updated so that it indicates functions registered as actions need to be exported, and the changelog updated to refer to the changes here and in <https://github.com/huderlem/porymap/pull/360>


GriffinRichards:
Subsumed by #390

PR #390, opened 2021-12-17
Expand API

GriffinRichards:
This adds ~60 new API functions, 7 new script callbacks, and new arguments to 5 existing API functions. All scripts made prior to these changes will continue to work as expected, as the new arguments are all optional and their default values will result in the same behavior as before. See the changes to the manual for descriptions of the new functions.

Notable changes:
- The overlay now has many "layers" (actually a `QMap` of layer numbers to overlay pointers) that can be manipulated independently. The layers are rendered in ascending order. This resolves my personal major issue with the overlay, which is that it was not previously possible to selectively remove anything; you had to clear everything and then reconstruct the overlay without whatever you wanted to remove. It is also possible to selectively move and hide layers.
- The API can now read and write metatile data, i.e. their tiles, label, and attributes. This is currently **unsafe** as it writes directly to the project and doesn't commit changes to the edit history. I take my cue from the `setPrimaryTilesetPalette` group of functions in assuming this is at least ok for now.
- There are now many more options for image creation by using `createImage`, as well as the ability to add tile and metatile images. Repeat image reading is faster now because Porymap will automatically cache images (unless asked not to).
- The cursor tile rectangle, player view rectangle, and map grid are now rendered manually in the foreground, as opposed to being rendered automatically as items on the items layer. This was to make sure they're always visible above the overlay, as most uses for the overlay (I think) look wrong when covering these UI elements. The implementation feels a bit workaround-y, and ideally I think the user would be able to control whether the overlay renders above or below these elements. Incidentally fixed a bug while doing this that allowed the tile cursor on the Events tab.
- Internally, the name `tile` is often used ambiguously to refer to tile ids and metatile ids as well as actual "tiles." Many were renamed `tileId` or `metatileId`.
- The new (internal) function `setBlockdata` is used in a few undo/redo functions instead of direct assignment so that `onBlockChanged` will get called for any blocks changed using the edit history.


src/ui/imageproviders.cpp
```diff
@@ -87,24 +98,24 @@ QImage getMetatileImage(
...
     }
     return tileset->tiles.value(local_index, QImage());
 }
 
-QImage getColoredTileImage(uint16_t tile, Tileset *primaryTileset, Tileset *secondaryTileset, QList<QRgb> palette) {
-    QImage tileImage = getTileImage(tile, primaryTileset, secondaryTileset);
+QImage getColoredTileImage(uint16_t tileId, Tileset *primaryTileset, Tileset *secondaryTileset, QList<QRgb> palette) {
+    QImage tileImage = getTileImage(tileId, primaryTileset, secondaryTileset);
     if (tileImage.isNull()) {
         tileImage = QImage(16, 16, QImage::Format_RGBA8888);
```
GriffinRichards, line 120: The fallback dimensions in this function are mistakenly (I believe) using the dimensions of an entire metatile rather than of a single tile. This can cause some unexpected behavior with `addTileImage`. If there wasn't a reason for this I'll change them to `8`.

src/core/map.cpp
```diff
@@ -310,6 +310,12 @@ void Map::setDimensions(int newWidth, int newHeight, bool setNewBlockdata) {
         setNewDimensionsBlockdata(newWidth, newHeight);
     }
 
+    int oldWidth = layout->width.toInt();
+    int oldHeight = layout->height.toInt();
+    if (oldWidth != newWidth || oldHeight != newHeight) {
+        Scripting::cb_MapResized(oldWidth, oldHeight, newWidth, newHeight);
```
huderlem, line 316: Should this occur after the actual layout dimensions are changed?  Seems like it's happening too early here, and can mislead the script.

src/mainwindow.cpp
```diff
@@ -1660,15 +1661,20 @@ void MainWindow::on_action_Save_triggered() {
     updateMapList();
 }
 
-void MainWindow::on_tabWidget_2_currentChanged(int index)
+void MainWindow::on_mapViewTab_tabBarClicked(int index)
 {
+    int oldIndex = ui->mapViewTab->currentIndex();
+    if (index != oldIndex)
+        Scripting::cb_MapViewTabChanged(oldIndex, index);
```
huderlem, line 1668: Same idea here--shouldn't the scripting hook be called after the actual index and switchover has completed?

src/scripting.cpp
```diff
@@ -159,6 +229,46 @@ QJSValue Scripting::dimensions(int width, int height) {
...
+}
+
 QJSEngine *Scripting::getEngine() {
     return instance->engine;
 }
+
+QImage Scripting::getImage(QString filepath) {
+    const QImage * image = instance->imageCache.value(filepath, nullptr);
+    if (!image) {
+        image = new QImage(filepath);
```
huderlem, line 270: What happens if the image filepath doesn't exist?  (Does it crash porymap?)

GriffinRichards: That's handled by `Overlay::addImage`:
```
if (image.isNull()) {
    logError(QString("Failed to load image '%1'").arg(filepath));
    return false;
}
```

GriffinRichards: (If the filepath doesn't exist it returns a null image, not a null pointer)

src/ui/graphicsview.cpp
```diff
@@ -16,15 +17,68 @@ void GraphicsView::mouseReleaseEvent(QMouseEvent *event) {
...
-
 void GraphicsView::moveEvent(QMoveEvent *event) {
     QGraphicsView::moveEvent(event);
     QLabel *label_MapRulerStatus = findChild<QLabel *>("label_MapRulerStatus", Qt::FindDirectChildrenOnly);
     if (label_MapRulerStatus && label_MapRulerStatus->isVisible())
         label_MapRulerStatus->move(mapToGlobal(QPoint(6, 6)));
 }
+
+void MapView::drawForeground(QPainter *painter, const QRectF&) {
+    foreach (Overlay * overlay, this->overlayMap)
```
huderlem, line 28: Is this guaranteed to iterate over the `QMap`'s keys in ascending order?  I'm having trouble finding documentation for that.

GriffinRichards: `QMap`s are sorted by key
https://doc.qt.io/qt-5/qmap.html#details

src/ui/mappixmapitem.cpp
```diff
@@ -95,6 +96,7 @@ void MapPixmapItem::shift(int xDelta, int yDelta, bool fromScriptCall) {
     }
 
     if (!fromScriptCall && map->layout->blockdata != oldMetatiles) {
+        Scripting::cb_MapShifted(xDelta, yDelta);
```
huderlem, line 99: Same thing here--call the script hook should probably be called after the map has actually been shifted.

GriffinRichards: The map shift occurs in the block above, so it already is.

GriffinRichards: Oh unless you mean it should also occur after the change to the edit history right below it

huderlem: I had misread the code, but yeah, I still think it should go after the edit history line.

huderlem:
The rest seems good to me

PR #405, opened 2022-01-21
Stop repeated parsing of event_objects.h and songs.h

GriffinRichards:
Any time the `OBJ_EVENT_GFX` or `MUS`/`SE` constants are needed Porymap parses them in their entirety from the file in the project. Now it only parses them once when the project is loaded and they're subsequently read from memory.


GriffinRichards:
Adding on: `loadEventPixmaps` was called for most event updates, and included parsing of files in `src/data/object_events`. It's been split into `readEventGraphics`, which is called once on project load and parses the files to determine the sprite data for each `OBJ_EVENT_GFX` constant, and `setEventPixmap`, which uses this data to set the sprite.

PR #408, opened 2022-02-03
Preserve unused metatile attributes

GriffinRichards:
All Gen III games have unused bits in their metatile attributes. Currently Porymap ignores these bits during reading/writing, meaning if users have set them for some purpose in their project then Porymap will clear them. Now Porymap will preserve these bits.

Also moves the logic for separating attributes to `metatile.cpp`, because the work to do this was being duplicated. Same for the fields of `Tile`.

### DIFF
--- a/src/project.cpp
+++ b/src/project.cpp
@@ -1034,21 +1034,12 @@ void Project::saveTilesetMetatileAttributes(Tileset *tileset) {
     QFile attrs_file(tileset->metatile_attrs_path);
     if (attrs_file.open(QIODevice::WriteOnly | QIODevice::Truncate)) {
         QByteArray data;
-
-        if (projectConfig.getBaseGameVersion() == BaseGameVersion::pokefirered) {
-            for (Metatile *metatile : tileset->metatiles) {
-                data.append(static_cast<char>(metatile->behavior));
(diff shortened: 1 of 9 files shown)


src/project.cpp
```diff
@@ -1571,13 +1558,8 @@ void Project::loadTilesetMetatiles(Tileset* tileset) {
...
             int index = i * (2 * 4 * num_layers);
             for (int j = 0; j < 4 * num_layers; j++) {
-                uint16_t word = data[index++] & 0xff;
-                word += (data[index++] & 0xff) << 8;
-                Tile tile;
-                tile.tileId = word & 0x3ff;
-                tile.xflip = (word >> 10) & 1;
-                tile.yflip = (word >> 11) & 1;
-                tile.palette = (word >> 12) & 0xf;
+                Tile tile(static_cast<unsigned char>(data[index++])
```
huderlem, line 1561: I'm worried these two `++` operator usages are actually undefined behavior, regarding order.  I don't know the answer, but it makes me squeamish.

GriffinRichards: Good call, my understanding is that unlike logical or there's no guaranteed order here.

PR #411, opened 2022-02-07
Better support for clone objects

### LINKED ISSUE #236: Split "in connection" events from Object Event
Something to do for the next major release (will require breaking changes). At the moment these are treated as the same event type both in porymap and in pokefirered. This results in unusual field usage (e.g. Trainer Type being the map number. See <https://github.com/pret/pokefirered/blob/6f05711b27ab9a1072e83c74791c6ef18177cac4/src/overworld.c#L411-L413>). `struct ObjectEventTemplate` should be split with a union to reflect the fields' actual usages, and then porymap changed to reflect that as well.

Name suggestion: Clone Object?

GriffinRichards:
**This is a breaking change for pokefirered that requires the commits in** https://github.com/pret/pokefirered/pull/484

"Clone objects" are a kind of object event treated specially in pokefirered. Clone objects store a map and event id (and a graphics id that isn't read), and when events are being loaded the event specified by this stored map/id is loaded instead. This is used to load objects visible in the connecting area of adjacent maps.

Porymap previously treated these events as regular object events, which resulted in their fields being displayed incorrectly (the targeted local id was shown as  the z coord, and the targeted map was shown as the trainer type and sight radius). Now they are their own event type on the Objects tab with properly labeled fields. The "in connection" field that was used to toggle whether to treat the object as a clone has been removed, and its corresponding config option has been renamed.

Additionally, the sprite displayed now reflects the graphics id and movement type of the targeted object, and double-clicking the event will open the targeted map with the targeted object selected. If the targeted map/object is invalid or is another clone then the sprite for graphics id 0 will be displayed.

Also included are some generic changes preemptively ported from the events API, notably constants for event group types.

Closes #236


src/core/event.cpp
```diff
@@ -182,81 +195,76 @@ int Event::getPixelY()
...
 };
 
-QMap<QString, bool> Event::getExpectedFields()
+const QStringList expectedSecretBaseFields = {
+    "type",
+    "elevation",
+    "secret_base_id",
+};
+
+QStringList Event::getExpectedFields()
```
huderlem, line 257: `QSet` is probably more appropriate than `QStringList`.

src/mainwindow.cpp
```diff
@@ -728,11 +718,14 @@ void MainWindow::openWarpMap(QString map_name, QString warp_num) {
         return;
     }
 
-    QList<Event*> warp_events = editor->map->events["warp_event_group"];
-    if (warp_events.length() > warpNum) {
-        Event *warp_event = warp_events.at(warpNum);
+    // Select the target event.
+    if (event_group != EventGroup::Warp && event_index)
```
huderlem, line 722: Can you add a comment here explaining why non-warp events use a 1-based `event_index`?

huderlem, line 722: Also, explicitly check that `event_index > 0`.

PR #412, opened 2022-02-08
Fix two cursor outline bugs

GriffinRichards:
- Normally the cursor tile and player view outlines disappear when the cursor exits the map. While clicking and dragging these outlines can be forced outside the map bounds on the right and bottom edges. This was because `updateCursorRectPos` was called before the check for map bounds.
- The cursor tile and player view outlines can be toggled with keyboard shortcuts. On the Metatiles tab they will update immediately, but on the Collision tab they will only update once the cursor enters a new space. This is because mainwindow was checking `has_mouse` on `map_item` but not `collision_item`.

### DIFF
--- a/src/editor.cpp
+++ b/src/editor.cpp
@@ -1004,14 +1004,10 @@ void Editor::setCursorRectVisible(bool visible) {
         ui->graphicsView_Map->scene()->update();
 }
 
-bool Editor::isWithinMap(int x, int y) {
-    return (x >= 0 && x < map->getWidth() && y >= 0 && y < map->getHeight());
-}
-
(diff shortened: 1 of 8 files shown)


src/editor.cpp
```diff
@@ -1004,25 +1004,33 @@ void Editor::setCursorRectVisible(bool visible) {
         ui->graphicsView_Map->scene()->update();
 }
 
+bool Editor::isWithinMap(int x, int y) {
```
huderlem, line 1007: nitpick: It feels like like bounds-checking method should belong in the `Map` class, rather than `Editor`.

PR #413, opened 2022-02-10
Add metatile label copy button

### LINKED ISSUE #395: Add the ability to copy a metatile label's full symbol to clipboard
A useful addition would be to have a little Copy button next to the metatile label field that copies the full symbol name to your clipboard.

For example, if I'm editing gTileset_Rustboro, the button when clicked on Door_Gray would copy `METATILE_Rustboro_Door_Gray` to my clipboard.

GriffinRichards:
Closes #395 

~~This needs an icon to match the style of other icons, currently it has a placeholder.~~


forms/tileseteditor.ui
```diff
@@ -271,42 +293,47 @@
...
-               <property name="clearButtonEnabled">
-                <bool>true</bool>
-               </property>
-              </widget>
-             </item>
-             <item row="1" column="2">
-              <widget class="QComboBox" name="comboBox_layerType"/>
-             </item>
-             <item row="8" column="0" colspan="3">
-              <widget class="NoScrollComboBox" name="comboBox_metatileBehaviors" native="true"/>
```
GriffinRichards, line 285: I also want to point out that QtCreator removed `native="true"` from `comboBox_metatileBehaviors`, not sure if that should be restored.

garakmon (commented):
If you copy a metatile label then paste it into the text box again, it keeps the prefixes. I wonder if they should be cleaned after pasting. Perhaps this is not worth it or even not desirable?

forms/tileseteditor.ui
garakmon: none of porymap's other comboboxes have this property so it's probably fine

GriffinRichards:
Hm, maybe? Assuming it makes sense to always check if a given label has the prefix and remove it, not just for pastes. I don't think copy/pasting the labels within the editor would be a common use case but it could happen if you wanted to relocate a metatile for instance, or if you were adding many new labels that were similar to an existing one.

PR #428, opened 2022-04-22
Improve Region Map Editor

garakmon:
- use a configurator to give users more options
- support firered (tilemaps, layout layers)
- add support for 4bpp and 8bpp tilemaps
- use new region map entries json format
- temporarily disabled city tilemap editor code


garakmon:
note: for best results, use the accompanying PRs in https://github.com/pret/pokefirered/pull/500 and https://github.com/pret/pokeemerald/pull/1651

PR #432, opened 2022-06-17
Add a color picker dialog to palette editor

### LINKED ISSUE #267: Suggestion: Allow manual entry of RGB Values in Tileset Palette Editor

### LINKED ISSUE #441: Manual missing documentation on Tileset Editor and Palette Editor
The Tileset Editor isn't explained beyond a brief intro under Navigation. The Palette Editor is never mentioned.

### LINKED ISSUE #443: Tileset Editor returning back to old metatile after saving
After you make/ edit an metatile in tileset editor and save it the tileset editor gonna refresh and instead of staying on tile you were editing it will return back to tile you were on while starting Tileset Editor

garakmon:
To use, open the palette editor and click the eyedropper button next to the color you wish to change.
Pressing the spacebar when your cursor is hovering over the desired color will pick that color.
Close the dialog if you want to abort.

*There is something wonky going on with my high-dpi display and magnifying text, but it's not a big deal in this application.

closes #267 
closes #441 
closes #443


src/ui/colorpicker.cpp
```diff
@@ -0,0 +1,81 @@
+#include "colorpicker.h"
+#include "ui_colorpicker.h"
+
+#include <QtWidgets>
+
+const int zoom_box_dimensions = 7;
```
GriffinRichards, line 6: Unused warning for this, was there a plan for it?

garakmon: ah whoops, I was hard coding while trying to figure out the pixelRatio thing and forgot to switch back
resolved now

PR #445, opened 2022-07-14
Add message boxes and user input windows to API

GriffinRichards:
The `QInputDialog` functions already handle clamping arguments to sane values.

Basic test script: https://mystb.in/HeadsetPreciousSeparately.cpp

### DIFF
--- a/docsrc/manual/scripting-capabilities.rst
+++ b/docsrc/manual/scripting-capabilities.rst
@@ -1229,9 +1229,74 @@ These are some miscellaneous functions that can be very useful when building cus
 
    :param string message: the message to log
 
-
 .. js:function:: map.error(message)
 
    Logs a message to the Porymap log file with the prefix ``[ERROR]``.
 
    :param string message: the message to log
+
+.. js:function:: map.showMessage(text, informativeText, detailedText)
+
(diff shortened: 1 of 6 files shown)


docsrc/manual/scripting-capabilities.rst
```diff
@@ -1229,9 +1229,74 @@ These are some miscellaneous functions that can be very useful when building cus
...
    :param string message: the message to log
 
-
 .. js:function:: map.error(message)
 
    Logs a message to the Porymap log file with the prefix ``[ERROR]``.
 
    :param string message: the message to log
+
+.. js:function:: map.showMessage(text, informativeText, detailedText)
```
huderlem, line 1238: These should probably exist under some `util`-like namespace instead of `map`, but I don't know if I care that much.

huderlem (approved):

docsrc/manual/scripting-capabilities.rst
GriffinRichards: Yeah probably. Same for Settings, and the Overlay functions to some degree. Some of the Tileset functions could be explained as applying to the tilesets for the current map, but some of them apply to all tilesets (e.g. `getMaxPrimaryTilesetTiles`). I just worry about losing the convenience of not having to think about what comes before the function name.

PR #446, opened 2022-07-27
Fix API crashes on bad palette ids / tile numbers

GriffinRichards:
The scripting API functions `createImage`, `addTileImage`, and both versions of `setMetatileTile` and `setMetatileTiles` can crash Porymap or give unexpected behavior if given a bad palette or tile number.


GriffinRichards:
Tacking on a fix for `map.setOverlayOpacity`. The opacity of the cursor rectangle and map grid were affected as well now that they're rendered after the overlay items.

GriffinRichards:
Merging with https://github.com/huderlem/porymap/pull/448

PR #448, opened 2022-08-03
Various minor API improvements

GriffinRichards:
- Collision is a 2 bit value, and `setBlock` / `setCollision` allow users to assign any value in this range. Porymap incorrectly renders collisions of 2 or 3 as the entire collision/elevation image. The games treat any nonzero collision value as "impassable," and now so does Porymap.
- Adds an overload of `setBlock` that takes the raw value given by `getBlock` or `onBlockChanged`, so you could more easily for instance disable drawing by doing:
```js
export function onBlockChanged(x, y, prevBlock, newBlock) {
    if (prevBlock.rawValue != newBlock.rawValue)
        map.setBlock(x, y, prevBlock.rawValue);
}
```
- Fixes a bug that allowed negative coordinates in `setBlock`
- The scripting API functions `createImage`, `addTileImage`, and both versions of `setMetatileTile` and `setMetatileTiles` can crash Porymap or give unexpected behavior if given a bad palette or tile number. These are fixed by making tile properties unsigned values of the correct width, and an explicit check in `Tileset::getPalette`.
- Fixes the unused layer of `METATILE_LAYER_MIDDLE_TOP` rendering with an incorrect palette
- Fixes a bug with `map.setOverlayOpacity` that caused some UI elements like the cursor rectangle to also change opacity.
- Adds `map.getMetatileAttributes` and `map.setMetatileAttributes` for reading/writing the raw metatile attributes value (giving users access to the unused attribute bits)

### DIFF
--- a/src/scripting.cpp
+++ b/src/scripting.cpp
@@ -239,17 +239,17 @@ QJSValue Scripting::position(int x, int y) {
 }
 
 Tile Scripting::toTile(QJSValue obj) {
-    if (!obj.hasProperty("tileId")
-     || !obj.hasProperty("xflip")
-     || !obj.hasProperty("yflip")
-     || !obj.hasProperty("palette")) {
-        return Tile();
-    }
     Tile tile = Tile();
-    tile.tileId = obj.property("tileId").toInt();
-    tile.xflip = obj.property("xflip").toBool();
-    tile.yflip = obj.property("yflip").toBool();
-    tile.palette = obj.property("palette").toInt();
+
+    if (obj.hasProperty("tileId"))
+        tile.tileId = obj.property("tileId").toInt();
+    if (obj.hasProperty("xflip"))
+        tile.xflip = obj.property("xflip").toBool();
+    if (obj.hasProperty("yflip"))
+        tile.yflip = obj.property("yflip").toBool();
+    if (obj.hasProperty("palette"))
+        tile.palette = obj.property("palette").toInt();
+
     return tile;
 }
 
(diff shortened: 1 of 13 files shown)


src/scripting.cpp
```diff
@@ -239,17 +239,17 @@ QJSValue Scripting::position(int x, int y) {
...
-     || !obj.hasProperty("palette")) {
-        return Tile();
-    }
     Tile tile = Tile();
-    tile.tileId = obj.property("tileId").toInt();
-    tile.xflip = obj.property("xflip").toBool();
-    tile.yflip = obj.property("yflip").toBool();
-    tile.palette = obj.property("palette").toInt();
+
+    if (obj.hasProperty("tileId"))
```
huderlem, line 244: double checking... are these calls to `.toInt()` and `.toBool()` still valid/functional after their type changes in `tile.h`?

src/ui/overlay.cpp
```diff
@@ -26,9 +26,11 @@ void OverlayImage::render(QPainter *painter, int x, int y) {
 void Overlay::renderItems(QPainter *painter) {
     if (this->hidden) return;
 
+    qreal oldOpacity = painter->opacity();
```
huderlem, line 29: There is a `painter->save()` and `painter->restore()` pair that pushes and pops the painter's state.  This is fine, too, though.

src/scripting.cpp
GriffinRichards: Yeah it's converting the JS property string to int/bool, and then converting (implicitly) to the bitfield types. `bool` to `uint16_t:1` is fine, `int` to `uint16_t:4` etc. might give a warning about dropping higher bits but that's what we want here anyway

PR #453, opened 2022-08-26
Support changes for dynamic warps

GriffinRichards:
See https://github.com/pret/pokeemerald/pull/1755

This is complete, but merging should probably wait until just before a new release as it's a breaking change for a common feature.


GriffinRichards:
Reimplementing this in #479

PR #456, opened 2022-08-31
Add scale to API's createImage, add getPorymapVersion

GriffinRichards:
Replaces the boolean arguments `xflip` and `yflip` in `createImage` with arguments for horizontal and vertical scale. This is a breaking change for the scripting API.

To help with future breaking changes this also adds `getPorymapVersion` to the API. I'm not happy with how it reads the version (by invisibly creating an "About" window and reading it from the UI text), feel free to change that. Test script: https://mystb.in/SpamJourneyAlloy

As with the last PR, skipped the changelog for now to avoid conflicts.


huderlem:
We can figure out a better way to encode Porymap's version some other time.  Manually bumping version numbers is error-prone, especially if there are multiple.

PR #470, opened 2022-10-04
Add Cut for metatiles in the Tileset Editor

GriffinRichards:
To complement the new Copy/Paste. Additionally moved all the instances of `getTripleLayerMetatilesEnabled() ? 12 : 8` to a single function.

### DIFF
--- a/src/ui/metatilelayersitem.cpp
+++ b/src/ui/metatilelayersitem.cpp
@@ -19,9 +19,9 @@ void MetatileLayersItem::draw() {
         QPoint(80, 16),
     };
 
-    int numTiles = projectConfig.getNumTilesInMetatile();
-    QPixmap pixmap(numTiles * 8, 32);
+    QPixmap pixmap(projectConfig.getNumLayersInMetatile() * 32, 32);
     QPainter painter(&pixmap);
+    int numTiles = projectConfig.getNumTilesInMetatile();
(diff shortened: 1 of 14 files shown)


src/ui/metatilelayersitem.cpp
```diff
@@ -19,11 +19,9 @@ void MetatileLayersItem::draw() {
         QPoint(80, 16),
     };
 
-    bool isTripleLayerMetatile = projectConfig.getTripleLayerMetatilesEnabled();
-    int width = isTripleLayerMetatile ? 96 : 64;
-    QPixmap pixmap(width, 32);
+    int numTiles = projectConfig.getNumTilesInMetatile();
+    QPixmap pixmap(numTiles * 8, 32);
```
huderlem, line 23: `numTiles * 8` doesn't make sense here, even though it's the correct value.  The pixmap is scaled to 2x here, so this should probably be rewritten as `(numTiles / 2) * 16`, or create a temporary variable called `numTilesWide = numTiles / 2`.

huderlem: Or maybe `projectConfig.getNumLayersInMetatile() * 32`

GriffinRichards: Ah, I had 8 pixels wide per tile in my head

PR #473, opened 2022-10-08
Support parsing tileset data in C

GriffinRichards:
- Maintains backwards compatibility, i.e. the current asm tileset data format can still be read/written, and will be used only (and automatically) if the C headers file is missing.
- Cuts down on some repeated parsing. Tileset labels are now only read from the headers file once, instead of every time a new tileset is loaded.
- The tileset header fields `isCompressed`, `callback`, and the padding are no longer read. Config options were added to stop Porymap from outputting `isCompressed` or `callback` in new C tileset headers.
- Adds 2 new parsing functions: `readCStructs`, a generalized version of the function used to parse `object_event_graphics_info.h`, and `readCIncbinArray`, to parse the arrays of `INCBIN`s used to define tileset palettes.

Note: This slightly disturbs Advance Map importing. The Advance Map importer assumes the tilesets are in the vanilla order, and uses this to set the default selected tileset when importing. Because tilesets are now parsed into a `std::map` by `ParseTopLevelObjects` they are instead in alphabetical order.


src/core/tileset.cpp
```diff
@@ -132,73 +126,152 @@ QList<QRgb> Tileset::getPalette(int paletteId, Tileset *primaryTileset, Tileset
...
         return false;
     }
-    QString dataString = "\n\t.align 2\n";
-    dataString.append(QString("gTilesetPalettes_%1::\n").arg(friendlyName));
-    for(int i = 0; i < Project::getNumPalettesTotal(); ++i) {
-        QString paletteString = QString("%1.gbapal").arg(i, 2, 10, QLatin1Char('0'));
-        dataString.append(QString("\t.incbin \"%1%2/%3/palettes/%4\"\n").arg(basePath, primaryString, friendlyName.toLower(), paletteString));
 
+    const QString tilesetDir = this->getExpectedDir();
+    const QString tilesPath = tilesetDir + "/tiles.4bpp.lz";
```
huderlem, line 183: I think this tiles filepath needs to check the compressed config option.

GriffinRichards: Porymap (currently at least) always outputs tileset paths compressed. We can change that if you want, idk how that would be presented in the UI. I don't think we should assume if `isCompressed` isn't being outputted that the user doesn't want it compressed

src/project.cpp
```diff
@@ -1907,46 +1930,68 @@ Project::DataQualifiers Project::getDataQualifiers(QString text, QString label)
...
+        while (iter.hasNext()) {
+            QRegularExpressionMatch match = iter.next();
+            insertTilesetLabel(match.captured("label"), match.captured("isSecondary"));
+        }
+        filename = asm_filename; // For error reporting further down
+    } else {
+        this->usingAsmTilesets = false;
+        const auto structs = parser.readCStructs(filename, "", Tileset::getHeaderMemberMap(this->usingAsmTilesets));
+        QStringList labels = structs.keys();
+        // TODO: This is alphabetical, AdvanceMap import wants the vanilla order in tilesetLabelsOrdered
```
huderlem, line 1979: Just checking, did you actually test the AdvanceMap import functionality with these changes, or is this hypothetical.  I haven't tested that in a long time, so I'm wondering if it still even works.

src/core/tileset.cpp
huderlem: I'm confused--isn't that exactly what the `isCompressed` option is used for?  It denotes if new tilesets have a compressed tiles image?

GriffinRichards: Porymap doesn't care if the image will be compressed, it just needs to know where the image is (and the image is always `.png`). Porymap also never edits existing tileset headers. When outputting a new tileset header, currently and in this PR it will always set `isCompressed` to `TRUE` and output a `.lz` file path. The config option is just to stop Porymap from outputting the `isCompressed` field if e.g. a user removed it from their struct

huderlem: Ok, I'm with you.  I was interpreting the `isCompressed` config option to mean "I don't want my tiles image to be compressed", rather than "compressed is not part of my tileset struct at all".

src/project.cpp
GriffinRichards: Hypothetical, based on my interpretation of this code
<https://github.com/huderlem/porymap/blob/2d87a6222a8cba1c5d5439a578235e4d04f6dfeb/src/core/mapparser.cpp#L78-L88>

huderlem: Ok, I'll test it just for sanity

src/core/tileset.cpp
GriffinRichards: Gotcha, I realize now I forgot to update the manual for those new config options (which should clear that confusion up)

src/project.cpp
huderlem: This is still working.  The user gets to choose the tileset in the import dialog anyway, so it's not a big deal.

PR #478, opened 2022-10-14
Add affine transformations and path drawing to the API

GriffinRichards:
- Adds 12 new API functions:
```
overlay.getHorizontalScale(layer = 0)
overlay.getVerticalScale(layer = 0)
overlay.setHorizontalScale(scale, layer)
overlay.setHorizontalScale(scale)
overlay.setVerticalScale(scale, layer)
overlay.setVerticalScale(scale)
overlay.getAngle(layer = 0)
overlay.setAngle(angle, layer)
overlay.setAngle(angle)
overlay.rotate(degrees, layer)
overlay.rotate(degrees)
overlay.addPath(xCoords, yCoords, borderColor = "#000000", fillColor = "", layer = 0)
```
- `addRect` and `addFilledRect` are combined into a single `addRect` function with new arguments
```
overlay.addRect(x, y, width, height, borderColor = "#000000", fillColor = "", rounding = 0, layer = 0)
```
- If an invalid color is given to an API function they will now log a warning and treat it as transparent. The empty string is silently treated as transparent.

Internally:
- Because the overlay is now setting its position with the painter's transform it no longer needs to pass its x and y to the items it renders.
- `OverlayRect` has been replaced by a generic `OverlayPath`. In theory this is faster as it now only has to create the rectangle once, but this is really a change of convenience so that we don't need separate classes for `addPath` and `addRect`.
- There is a currently an unused `addPath` overload that takes a single list of x/y pairs. I haven't been able to test that yet.
- Removed some defaults for overlay functions that always get their arguments provided anyway by the API function calling them


docsrc/manual/scripting-capabilities.rst
```diff
@@ -1010,6 +1010,79 @@ All overlay functions are callable via the global ``overlay`` object.
...
+   :param number scale: the scale to set
+   :param number layer: the layer id
+
+.. js:function:: overlay.setVerticalScale(scale)
+
+   This is an overloaded function. Sets the vertical scale of all active overlay layers. Layers that have not been used yet will not have their scale changed. ``1.0`` is normal size.
+
+   :param number scale: the scale to set
+
+.. js:function:: overlay.getAngle(layer = 0)
```
huderlem, line 1053: These `angle`-named functions should use the word `rotation` instead.  I.e. `getRotation`, `setRotation`.

docsrc/manual/scripting-capabilities.rst
```diff
@@ -1123,30 +1196,31 @@ All overlay functions are callable via the global ``overlay`` object.
...
    :param number width: the pixel width of the rectangle
    :param number height: the pixel height of the rectangle
-   :param string color: the color of the rectangle. Can be specified as "#RRGGBB" or "#AARRGGBB". Defaults to black.
+   :param string borderColor: the color of the rectangle's border. Can be specified as ``"#RRGGBB"`` or ``"#AARRGGBB"``. Defaults to black.
+   :param string fillColor: the color of the area enclosed by the rectangle. Can be specified as ``"#RRGGBB"`` or ``"#AARRGGBB"``. Defaults to transparent.
+   :param number rounding: the percent degree the corners will be rounded. ``0`` is rectangular, ``100`` is elliptical. Defaults to ``0``
    :param number layer: the layer id. Defaults to ``0``
 
-.. js:function:: overlay.addFilledRect(x, y, width, height, color = "#000000", layer = 0)
+.. js:function:: overlay.addPath(xCoords, yCoords, borderColor = "#000000", fillColor = "", layer = 0)
```
huderlem, line 1216: Don't forget to add docs for the 2d-array version of `addPath()`.

docsrc/manual/scripting-capabilities.rst
GriffinRichards: Yeah I had them that way originally but I renamed them because I thought it was less accurate. It probably makes more sense for them to use a cohesive name, I'll revert it

PR #480, opened 2022-10-19
Display values for missing items in combo boxes

GriffinRichards:
For the `Movement`, `Trainer Type`, `Layer Type`, `Encounter Type`, and `Terrain Type` combo boxes, display the given value if it doesn't exist as one of the items.

Fixes a regression for this behavior on the first two, which is relevant in vanilla for e.g. the Seafoam Island boulders. These events use Trainer Type to store a flag value.

As I'm on Qt 5.14 I can't test the changes to the Tileset Editor boxes. If you run this script
```js
export function onProjectOpened(projectPath) {
    utility.registerAction("runTest", "Run Test");
}
export function runTest() {
    map.setMetatileAttributes(1, 0xFFFFFFFF);
}
```
and select metatile 1 in the Tileset Editor, you should see values displayed in the Metatile Behavior, Layer Type, and (if pokefirered) the Encounter and Terrain Type combo boxes, as opposed to empty boxes or text.

### DIFF
--- a/include/ui/noscrollcombobox.h
+++ b/include/ui/noscrollcombobox.h
@@ -10,6 +10,7 @@ class NoScrollComboBox : public QComboBox
 public:
     explicit NoScrollComboBox(QWidget *parent = nullptr);
     void wheelEvent(QWheelEvent *event);
+    void setTextItem(const QString &text);
 
 private:
 };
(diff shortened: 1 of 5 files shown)


huderlem (approved):
Confirmed this works with Qt 6

PR #492, opened 2022-11-23
Random Speed Improvements

garakmon:
I've found some places where porymap could be faster so I'll leave them here.


src/core/parseutil.cpp
```diff
@@ -51,6 +58,10 @@ QString ParseUtil::createErrorMessage(const QString &message, const QString &exp
 }
 
 QString ParseUtil::readTextFile(const QString &path) {
+    if (fileCache.contains(path)) {
+        return fileCache[path];
+    }
```
GriffinRichards, line 63: This introduces a bug when `readTextFile` is used on a file that has changed, because it will instead return the file in its original state. Example: Open the palette editor, change a color, save and exit the palette/tileset editor, change to a different tileset, change back to the previous tileset, re-open the palette editor, the color change has been lost

GriffinRichards:
I haven't tested them, but after the recent changes to the file cache I believe at least the following files will still have problems (all read by `Project::readTilesetPaths`):
- `graphics.inc` / `graphics.h`, written in `Tileset::appendToGraphics`.
- `metatiles.inc` / `metatiles.h`, written in `Tileset::appendToMetatiles`.
- `headers.inc`, written in `Tileset::appendToHeaders`. `headers.h` is also written here, but it's fine because it's read by the C parser and doesn't use your file cache at all.

GriffinRichards:
Another: the log file is written without updating the cache by `log`, and read using the cache in `ParseUtil::textFileLineCount`

GriffinRichards (approved):

PR #512, opened 2023-01-26
Improve layout parser warnings/errors

GriffinRichards:
Got tired of the low readability on these warnings.
Changes:
- Don't specify the root in the filepath if the error wasn't opening the file
- Give the name of the layout, instead of the index number
- Log warnings for duplicate layout entries

Old example:
```
<time>  [WARN] Invalid layout 'border_height' value '0' on layout 76 in /Users/GriffinR/Desktop/pokefirered/data/layouts/layouts.json. Must be greater than 0. Using default (2) instead.
```

New example:
```
<time>  [WARN] Invalid 'border_height' value '0' for LAYOUT_RS_SAFARI_ZONE_REST_HOUSE in data/layouts/layouts.json. Must be greater than 0. Using default (2) instead.
```

### DIFF
--- a/src/project.cpp
+++ b/src/project.cpp
@@ -410,11 +410,12 @@ bool Project::readMapLayouts() {
     mapLayouts.clear();
     mapLayoutsTable.clear();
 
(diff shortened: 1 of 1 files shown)


garakmon (commented):
I guess the `new` layout should be `delete`d at every early return

PR #515, opened 2023-02-02
Separate Layouts and Maps

### LINKED ISSUE #181: Edit Map Groups
It might be nice to say, change a map's group, rename map groups, rearrange maps in groups, etc.

### LINKED ISSUE #182: Open / edit layouts with no associated map
At the moment, to open/edit layouts that don't have an associated map you have to create a dummy map, which clutters the project and is tedious to do for multiple layouts. The ability to open/edit these layouts without having to do so would be convenient

### LINKED ISSUE #334: Changes to layouts.json can be saved separately from saving the actual map.bin, which can lead to incorrect sizes
Steps to repro:

1. Change size of layout on map 1; don't save
2. Go to map 2, change layout, and save
3. Observe that in layouts.json, the size of the layout for map 1 has changed
4. Cancel changes to layout 1 by reloading project, or exiting Porymap
5. Observe that map 1 has loaded the bin file incorrectly due to the layout being the wrong size

### LINKED ISSUE #477: Layout reloading can overwrite unsaved changes
Layout block data will be re-read from the project file if a newly opened map shares a layout with a previously opened map. This can overwrite any unsaved changes the user has made to the layout

garakmon:
Feature-complete. In need of cleanup and testing.

Internally removes Layout related code from the Map class.
Creates a new layout-only edit mode.
Adds ability to modify map groups.

closes #182 
closes #477 
closes #181 
closes #334 

TODO:
- [x] Edit history in timeplapse replays both layouts and events ~~in order of edits made~~
- [x]  Add new map group & map section & layout without map + delete all above + #550 +rename
- [x]  `ScriptUtility::setMainTab(int index)` consider edit mode
- [ ]  reassigning layouts silences "Setting Map" message log
- [x] toggle view on empty folders in mapList (map sections, groups, layouts)
- [x] number map sections?
- [x] fix all scripting API functions
- [x] rearrange map groups?
- [x] drop `gMapGroup_` prefix requirement for map groups
- [x] use `NewMapPopup::initUi` where needed
- [x] [BUG] switching to "Connections" tab disables map edits permanently
- [x] Fix loading recent map except it's a layout
- [ ] ~~Rearrange and rename map sections?~~


GriffinRichards:
Alright I'm nearly done adding to this. The main outstanding thing I'd like to fix is how we're reading/writing the `constants_region_map_sections` file. If I can update the project repos to generate that file using `json_region_map_entries` it should be a much simpler process.

There's a few bugs I'd like to address after that (mainly that resizing the map with the scripting API doesn't quite work right now), and then the rest of the `TODO` comments I added I'll leave to be addressed after this PR is merged:
- I have some quibbles with the New Map/Layout dialogs, but I have a bunch of incoming changes for the New Map dialog so I'll address those then.
- Once the config updates are finished, the state of the new toggleable buttons in the map list tool bar should be saved in the config.
- I left a bunch of comments related to the lack of support for deleting maps/layouts, that's obviously a bigger thing that can be addressed some other time.
- Creating a new map automatically saves the project. It should allow users to close and discard the changes instead (this was true before the PR, though the behavior is changed slightly)
- The API should be updated to have some functionality for the new layout features (e.g. a callback when a layout is opened or swapped)

PR #584, opened 2024-01-11
Update Custom Attributes table

### LINKED ISSUE #526: Automatically add custom fields
Often when someone adds a Custom Field to a map header or event they want it to apply to many/all maps or events. It'd be nice if Porymap supplied an option to make a Custom Field "universal", adding it to existing maps/events and continuing to add it when new ones are created (perhaps notifying the user that its current value will become that field's default value).

GriffinRichards:
- Key/value pairs added to Custom Attribute tables can now be set as defaults, meaning when a new map/event is added that key/value pair will be added to its table automatically. These are stored in the config
- Numbers in the Custom Attribute tables are now displayed with a value spinner
- The "Type" column is removed from the table, as data types are now evident when you look at the value cell. Data types can still be specified when you add Custom Attributes
- Adding keys that will collide with external keys is now prevented with an error message, rather than allowed to silently fail
- Adding empty keys is now prevented, rather than allowing them to disappear on reload
- Editing the Custom Attributes table for an Event now marks the map as unsaved

Internally, rewrote a bunch of `customattributestable` because it was handled unusually (the class was only really for Events, and it shared static functions with the Header tab's version of the table; now they fully share the class)

Closes #526 

TODO:
- [x] Fix the table briefly squishing visually when the first cell is added
- [x] Fix a resizing issue when the table is first displayed
- [ ] Add right-click context menu (or a new button?) for setting/unsetting defaults
- [ ] Add edit history to the table
- [ ] Handle interaction with `Project Settings` window
- [ ] Take control of the Delete key when the table has focus?
- [ ] Investigate whether it's reasonable to support adding default attributes to all existing maps/events as well, rather than just new ones
- [ ] Update manual


GriffinRichards:
Closing in favor of #669

PR #588, opened 2024-01-18
Stop changing settings if their features fail to load

GriffinRichards:
The settings for enabling Wild Encounters, Weather Trigger Events, and Secret Base Events may be changed if something they rely on fails to load. This can be confusing as users might not realize Porymap automatically changed their settings, and the feature won't resume working once the problem is resolved.

Now these features are only disabled for the session in which they failed to load.


GriffinRichards:
Updated to fix a crash that can occur if Secret Base or Weather Trigger events are disabled, but present in the project. In this case it was possible to select one of these events and crash Porymap by trying to add more with the Pencil tool.

This was fixed by relaxing the requirements to open a project. Prior to this Porymap required a project to have constants used by a number of different dropdowns. Now the dropdowns can be left empty, and Porymap will initialize new instances of these dropdowns with a "0".

After this PR, the requirements to open a project are just having valid data for at least 1 primary and secondary tileset, 1 layout, and 1 map + map group + map section (and this last requirement could probably be relaxed after #515).

PR #590, opened 2024-01-21
Add update promoter

GriffinRichards:
Adds a "Check for Updates" menu action that shows a dialog to check whether a new Porymap release is available. If one is available it will present a changelog from all release versions since the host version and provide a link to the downloads page. This also runs automatically on startup, which can be turned off under Preferences.

dialog

### DIFF
--- a/forms/aboutporymap.ui
+++ b/forms/aboutporymap.ui
@@ -52,9 +52,6 @@
         <pointsize>12</pointsize>
        </font>
       </property>
-      <property name="text">
-       <string>Version 5.3.0 - January 15th, 2024</string>
-      </property>
       <property name="alignment">
        <set>Qt::AlignCenter</set>
       </property>
--- a/forms/mainwindow.ui
+++ b/forms/mainwindow.ui
@@ -1715,7 +1715,7 @@
                              <x>0</x>
                              <y>0</y>
                              <width>100</width>
-                             <height>30</height>
+                             <height>16</height>
                             </rect>
                            </property>
                            <property name="sizePolicy">
@@ -1809,7 +1809,7 @@
                              <x>0</x>
                              <y>0</y>
                              <width>100</width>
-                             <height>30</height>
+                             <height>16</height>
                             </rect>
                            </property>
                            <property name="sizePolicy">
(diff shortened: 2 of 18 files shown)


GriffinRichards:
Some notes to self for updating this PR after reading https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api
- Update error handling to follow redirects (`QNetworkReply::redirected`)
- Record and use data (`QNetworkRequest::ETagHeader` and `QNetworkRequest::IfNoneMatchHeader`) to avoid hitting rate limits (I don't think Porymap has the volume for this to be a big issue, but regardless), and limit use of the `Retry` button.
- Handle GitHub's own rate limit headers

GriffinRichards:
Updated to address above points (cache results from server, respect rate limiting) as well as some other improvements like handling pagination.

JakobDev:
Could you add a Build flag to disable the Update check?

GriffinRichards:
I'm satisfied with users being able to turn it off under `Preferences`. If the network connection fails for any reason everything else continues to function as before. Is this a request related to flatpak?

JakobDev:
> Is this a request related to flatpak?

Yes, it would not make much sense to offer binary Downloads if the Program is installed through Flatpak. The Releases section of GitHub also don't offer any Linux binary at all.

GriffinRichards:
How does it relate to this PR, I mean. Why do you want a build flag to turn off network functions?

JakobDev:
> how would that be presented to flatpak users?

It depends on the Distro how Flatpak are updated in a graphical way. For the command line you can just run `flatpak update`. Most Distros will show a Notifications if Software can be updated.

I usually Update the Flatpak the same Day a new version of Porymap comes out. Flathub also has a publishing delay of a few hours. If a User starts Porymap directly after a Update comes out, he will be prompted to update but a new Version of the Flatpak is not available yet.

There are also other Packages of Porymap besides the Flatpak e.g. this [AUR packages](https://aur.archlinux.org/packages/porymap-git).

On Linux it is usual that Software don't check for Updates and let the System do the work.

GriffinRichards:
Makes sense, thanks! Updated, the feature is now exclusive to the platforms we provide binaries for (Windows and macOS).

PR #609, opened 2024-09-09
Include latest commit hash in version info

GriffinRichards:
It would sometimes be useful to get the hash of the latest commit Porymap was compiled with when getting bug information from users.

The displayed commit is only accurate once the Makefile is regenerated from `porymap.pro` and `aboutporymap.cpp` is recompiled, but this is also true for the version info (except it's `mainwindow.cpp` instead) so I'm not super worried about it.

about

- [x] ~Need to investigate what's happening in the GitHub actions for this~ (the commit hash for Linux and macOS was not in the repo, but that's because it was the hash of the pull request's test merge. This won't be a problem once it's actually merged)
- [x] Investigate why the windows static release build isn't getting a commit hash


GriffinRichards:
Updated to...
- Correctly find the latest commit hash on Windows
- Add the Qt version Porymap was compiled with
- Remove the changelog (I'm not sure this is the place to put it? It's a ton of information for what should be a concise window. It's still available via the documentation link, and via the update promoter when a new release is available, and if we really want the changelog always available in-app we could add a separate menu action for it).

PR #700, opened 2025-03-08
Use a QImage to Render Overlay Layers

garakmon:
Motivation here is to be able to control transparency on a layer wide basis and not OverlayItem basis.

Needs some work to investigate why certain scripts are made slower (eg Porymap-Animation), and how to update them.

Example script before and after
-------------------------------

Before:
Screenshot 2025-03-08 at 12 24 41

After:
Screenshot 2025-03-08 at 12 21 47

```js

var toggledOn = false;

export function show_player_visible_tiles(toggle = true) {
    let mapWidth = map.getWidth();
    let mapHeight = map.getHeight();

    const viewWidth = 15;
    const viewHeight = 10;

    let widthMargin = (viewWidth - 1) / 2 * 16;
    let heightMargin = (viewHeight - 1) / 2 * 16;

    if (toggle) {
        toggledOn = !toggledOn;
    }
    if (!toggledOn) {
        overlay.clear();
        return;
    }

    overlay.clear();

    for (let i = 0; i < mapWidth * mapHeight; i++) {
        let x = i % mapWidth;
        let y = Math.floor(i / mapWidth);
        if (!map.getCollision(x, y)) {
            overlay.addRect(x * 16 - widthMargin, y * 16 - heightMargin, viewWidth * 16, viewHeight * 16, "#00ff00", "#00ff00");
        }
    }

    overlay.setOpacity(30);
    overlay.show();
}

export function onProjectOpened(projectPath) {
    utility.registerToggleAction("show_player_visible_tiles", "Toggle Visibility Overlay", "Ctrl+/");
}

export function onBlockChanged(x, y, prevBlock, newBlock) {
    // Need to redraw visible tiles
    show_player_visible_tiles(false);
}

export function onMapOpened(mapName) {
    overlay.clear();
    if (toggledOn) {
        show_player_visible_tiles(false);
    }
}

```


GriffinRichards:
> Needs some work to investigate why certain scripts are made slower 
I think it's because for every overlay layer you're creating you're filling a `QImage` the size of the entire scene.

garakmon:
> > Needs some work to investigate why certain scripts are made slower
> 
> I think it's because for every overlay layer you're creating you're filling a `QImage` the size of the entire scene.

Yeah, is each metatile frame in its own layer? I haven't looked closely at the script yet.

GriffinRichards:
There's a few optimizations beyond that but yes on some maps it can use thousands of layers (which should be fine, that works pretty efficiently right now but it will fully freeze Porymap on some maps with this change). I'm not sure why a given layer should assume it will use the entire scene.

garakmon:
I guess I didn't understand what layers were for in the context of overlays. In my head they were analogous to metatile layers, and it would make sense that each frame of the animation is its own layer. But I can at least see how that could be inefficient, especially when only a subsection of the layer needs to be updated.
Fwiw the reason each overlay layer is for now the size of the entire map+borders is that I needed to draw in negative map coordinates, and I didn't feel like experimenting with QImage::scaled.

GriffinRichards:
> I guess I didn't understand what layers were for in the context of overlays.

When you add an item to the overlay you aren't given a way to refer back to that item, so layers were a way to A: refer back to them so you could change the item without clearing the overlay and recreating it, and B: group things together to make changes to multiple items. If there were a way to address changes to a specific item within a layer then something like the animation script wouldn't need to use nearly as many layers, but I still think it should be reasonably efficient to have a lot of layers.

PR #734, opened 2025-05-21
Fix some issues with config loading

### LINKED ISSUE #649: Porymap sometimes resets project config and deletes prefabs when opened
## Description

Sometimes when opening Porymap, the project config file is changed back to the default parameters. This occurs in a seemingly random fashion, but I have noticed it happening more when I first open Porymap after some time not using it (eg. in the morning).
If I change the config file back to my custom parameters, close and reopen Porymap, the file is not overwritten.

## Reproduction

- Open `porymap.project.cfg` with some custom parameters
- Open Porymap
- Check `porymap.project.cfg` again: all values have been changed back to the default parameters

## Setup

**OS:** Windows 11 Pro (64 bits), version 23H2
**Porymap version:** 5.4.1 - Mar 21 2024
**Decomp project:** `pokeemerald-expansion`, `upcoming` branch (last updated 02 Dec 2024 ~1.9.3)
**WSL setup:** Using WSL 2. `pokeemerald-expansion` is located on the linux filesystem, Porymap is located on the Windows filesystem

GriffinRichards:
- Fix handling of empty config files. Previously, an empty project config would be initialized with Emerald's default settings. Now it will be treated the same as if the file didn't exist (i.e., it will get the base game version, then initialize settings based on that version).
- Opening a project is now aborted if a user/project config exists but fails to open. Previously it would open the project anyway and initialize settings with the defaults, which could potentially overwrite the existing settings.
- `getConfigFilepath` was recreating the config filepath every time it was called, which included (for some configs) checking the file system to see if the root directory exists. Now the config filepath is only created once.

Closes #649

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -29,11 +29,21 @@ class KeyValueConfigBase
 {
 public:
     bool save();
-    void load();
-    virtual ~KeyValueConfigBase();
+    bool load(const QString &dir = QString());
+
+    void setRoot(const QString &dir);
+    QString root() const { return m_root; }
+    QString filepath() const { return m_filepath; }
+    QString filename() const { return m_filename; }
+
+    explicit KeyValueConfigBase(const QString &filename)
+        : m_root(QString()),
+          m_filename(filename),
+          m_filepath(filename)
+    { };
+    virtual ~KeyValueConfigBase() {};
     virtual void reset() = 0;
 protected:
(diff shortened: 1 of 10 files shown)


GriffinRichards:
(Ignoring the failed check, the Windows workflow is still broken)

PR #759, opened 2025-07-22
Fixed INSTALL.MD

ActuallyTheOwner:
Add Gentoo and Debian to the INSTALL.MD
Clarify steps for Linux Mint
Minimum windows version is windows 7 from what I can tell, this should be stated for older machines.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,16 +1,14 @@
 # Installation
 
-**Note**: For Windows and macOS, installation is not required to use Porymap. You can download the latest release to begin using Porymap immediately.
+For installation, Porymap requires tools for Qt 5.14.2 & C++11 or newer.
+
+**Note**: For Windows 7 or newer, and macOS, installation is not required to use Porymap. You can download the latest release to begin using Porymap immediately.
 
  - [Download Porymap for Windows](https://github.com/huderlem/porymap/releases/latest/download/porymap-windows.zip).
  - [Download Porymap for macOS latest (arm)](https://github.com/huderlem/porymap/releases/latest/download/porymap-macos-latest.zip).
  - [Download Porymap for macOS 13 (intel)](https://github.com/huderlem/porymap/releases/latest/download/porymap-macos-13.zip).
 
-
-For installation, Porymap requires Qt 5.14.2 & C++11.
-
 ## macOS
-
 The easiest way to get Qt is through [homebrew](https://brew.sh/). 
 Once homebrew is installed, run these commands in Terminal:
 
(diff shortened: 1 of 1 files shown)


ActuallyTheOwner:
I forgot to add anything on checking cpu threads for the make command
IE make -j12 for the AMD 5600x

Don't want to go out of scope though
Generally threads are double the core count, but not for some older CPU

ActuallyTheOwner:
Note clang does not build
I am unsure if there should be an addition to the installmd for this, or a patch for clang

INSTALL.md
```diff
@@ -1,16 +1,14 @@
 # Installation
 
-**Note**: For Windows and macOS, installation is not required to use Porymap. You can download the latest release to begin using Porymap immediately.
+For installation, Porymap requires tools for Qt 5.14.2 & C++11 or newer.
```
GriffinRichards, line 3: Please restore this line to its original wording and position (people coming from the pret repos are used to heading straight to `INSTALL`, so it's important for the first thing they see in here to be the "you don't need to do this" note, and Qt itself is required, not tools for Qt).

INSTALL.md
```diff
@@ -30,37 +28,52 @@ make
...
-
+## Windows 7+
 Install [Qt development tools](https://www.qt.io/download-qt-installer), and use Qt Creator, the official Qt IDE, for development purposes.
 
-## Ubuntu
 
-You need to install Qt. The minimum supported version is currently Qt 5.14.2. You can check your Qt version
-with `qtdiag` or `qmake --version`.
+## Gentoo Linux
+You need to install Qt dev tools. You can set version of your Qt packages with flags such as `qt6`.
```
GriffinRichards, line 36: It could be that I'm missing context here, but this seems unclear (flags provided to what? and how?)

INSTALL.md
```diff
@@ -30,37 +28,52 @@ make
...
-# if your distro does not have qt6-declarative-dev, try sudo apt-get install qtdeclarative5-dev
+sudo emerge dev-util/kdevelop
+sudo emerge --depclean
+qmake6
+make
+./porymap
+```
 
-git clone https://github.com/huderlem/porymap
-cd porymap
```
GriffinRichards, line 47: Please include these steps. They're often assumed, but we get many users for whom this process is unfamiliar (and the steps were added recently because people had missed them).

INSTALL.md
```diff
@@ -30,37 +28,52 @@ make
...
+## Debian /Linux derivatives (Non Ubuntu exception to Mint)
+You need to install Qt6.
+Qt 5.14.2 is porymap's minumum supported qt version qt 6 is recomended since Debian Bookworm
```
GriffinRichards, line 50: These are conflicting statements (and the second line is a run-on)

INSTALL.md
```diff
@@ -30,37 +28,52 @@ make
...
+qmake
+make
+./porymap
+```
 
-git clone https://github.com/huderlem/porymap
-cd porymap
+## Ubuntu derivatives (Non Mint)
+You need to install Qt. The minimum supported version is currently Qt 5.14.2. You can check your Qt version
+with `qtdiag`.
```
GriffinRichards, line 72: Should be the same as above, "with `qtdiag` or `qmake --version`".

INSTALL.md
```diff
@@ -1,16 +1,14 @@
 # Installation
 
-**Note**: For Windows and macOS, installation is not required to use Porymap. You can download the latest release to begin using Porymap immediately.
+For installation, Porymap requires tools for Qt 5.14.2 & C++11 or newer.
+
+**Note**: For Windows 7 or newer, and macOS, installation is not required to use Porymap. You can download the latest release to begin using Porymap immediately.
```
GriffinRichards, line 5: > For Windows 7 or newer

Did you verify this? The Windows release is compiled using Qt6, which isn't officially supported for Windows 7, so I wouldn't be surprised if it doesn't work. I would otherwise keep this as `For Windows and macOS`.

GriffinRichards (changes requested):
Thanks for adding additional installation instructions!

> Note clang does not build
> I am unsure if there should be an addition to the installmd for this

Hard to say without seeing your build output. I compile Porymap regularly with clang 15.0.0 on macOS.

INSTALL.md
ActuallyTheOwner: My mistake, the Qt6 for Windows 7 is a backport and not official https://github.com/crystalidea/qt6windows

INSTALL.md
```diff
@@ -30,37 +28,52 @@ make
...
+qmake6
 make
 ./porymap
 ```
 
-## Arch Linux
-
+## Arch Linux derivatives 
 You need to install Qt. You can check the version of your Qt packages with `qtdiag` or `qmake --version`.
+Qt 5.14.2 is porymap's minumum supported qt version.
```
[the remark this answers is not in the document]
ActuallyTheOwner: Bit of a copy paste issue I did not catch

PR #772, opened 2025-11-10
Added expansion multi-region support

Starkiller4011:
This is a change to support moving the region information for map sections into the map header information to enable multi-region support in expansion. The changes were done in a way that has no impact on projects that do not use the feature as the drop-down menu will only be enabled and visible in projects that are using it and the map.json will only have the new field written to if the project is using it and the map has had it's region set to a value other than REGION_NONE. These changes are to support the following pull request: https://github.com/rh-hideout/pokeemerald-expansion/pull/8160


GriffinRichards:
Hi, thanks for the PR!

There are a good number of unnecessary changes in here, I assume because this was mostly a copy-paste of how Porymap handles `region_map_sections`. Those changes aside, I think the field that this adds would be better supported as a generic improvement to `Custom Attributes`. Currently, an expansion user could equally go to the `Header` tab, click `Add` under `Custom Attributes`, and enter their region there
region
and they'd then be able to edit it normally on the `Header` tab
region2

The two drawbacks I see to this approach are that a user has to add the field themselves on each new map they want to use it for, and they don't have the options available in a dropdown. Fixing both of these drawbacks are already planned additions to `Custom Attributes`. Given there's no novel features here (it's just a straightforward editable field) I think your field is better served by those planned additions.

PR #807, opened 2026-07-19
Redesign species distribution chart

cawtds:
Redesigns the species distrubtion chart showing 1 horizontal bar per species. This allows for y-axis labels to include species + icon, making it visually easier to connect a species with its bar.
Example:  
image

### DIFF
--- a/include/ui/wildmonchart.h
+++ b/include/ui/wildmonchart.h
@@ -77,11 +77,13 @@ private:
 
 #else
 
+class Project;
+
 class WildMonChart : public QWidget
 {
     Q_OBJECT
 public:
-    explicit WildMonChart(QWidget *, const EncounterTableModel *) {};
+    explicit WildMonChart(QWidget *, const EncounterTableModel *, Project *) {};
     ~WildMonChart() {};
 
 public slots:
--- a/include/ui/wildmonchart.h
+++ b/include/ui/wildmonchart.h
@@ -8,6 +8,8 @@
 #if __has_include(<QtCharts>)
 #include <QtCharts>
 
+class Project;
+
 namespace Ui {
 class WildMonChart;
 }
@@ -16,7 +18,7 @@ class WildMonChart : public QWidget
 {
     Q_OBJECT
 public:
-    explicit WildMonChart(QWidget *parent, const EncounterTableModel *table);
+    explicit WildMonChart(QWidget *parent, const EncounterTableModel *table, Project *project);
     ~WildMonChart();
 
 public slots:
@@ -27,6 +29,7 @@ public slots:
 private:
     Ui::WildMonChart *ui;
     const EncounterTableModel *table;
+    Project *project = nullptr;
 
     QStringList groupNames;
     QStringList groupNamesReversed;
--- a/include/ui/wildmonchart.h
+++ b/include/ui/wildmonchart.h
@@ -5,10 +5,10 @@
 
 #include <QWidget>
 
+class Project;
+
 #if __has_include(<QtCharts>)
 #include <QtCharts>
-
-class Project;
 
 namespace Ui {
 class WildMonChart;
@@ -49,8 +49,6 @@ private:
     typedef QMap<QString, Summary> GroupedData;
 
     QMap<QString, GroupedData> speciesToGroupedData;
-    QMap<QString, QColor> speciesToColor;
-
 
     QStringList getSpeciesNamesAlphabetical() const;
     double getSpeciesFrequency(const QString&, const QString&) const;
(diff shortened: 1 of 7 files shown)


GriffinRichards:
The chart appears to be broken on encounter sets with multiple groups

CleanShot 2026-07-19 at 09 59 54@2x

Here's what that data looks like on your branch
CleanShot 2026-07-19 at 09 59 00@2x

Here's what it looks like currently
CleanShot 2026-07-19 at 11 03 54@2x


You'll also need to make sure those two Qt5 Linux jobs pass. In particular, one is checking to make sure people can build without having installed any optional modules like QtCharts.

src/ui/specieschartview.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+void SpeciesChartView::setSpecies(const QStringList &species)
+{
+    m_speciesEntries.clear();
+
+    const QString speciesPrefix = projectConfig.getIdentifier(ProjectIdentifier::define_species_prefix);
+
+    QFontMetrics fm(font());
+
+    m_maxTextWidth = 0;
```
GriffinRichards, line 21: We should have an upper limit to how much text we display here so we don't squish the chart, and elide names that are longer. Expansion in particular has some quite long species names. As an extreme example of what happens to the new chart

Image

src/ui/wildmonchart.cpp
```diff
@@ -223,64 +228,85 @@ bool WildMonChart::usesGroupLabels() const {
...
+    int maxTextWidth = 0;
+    for (const QString &category : categories) {
+        maxTextWidth = std::max(maxTextWidth, fm.horizontalAdvance(category));
+    }
+
+    int labelWidth = maxTextWidth + SpeciesChartView::IconSize + SpeciesChartView::Spacing;
+    int leftMargin = labelWidth + 2 * SpeciesChartView::Padding;
+    chart->setMargins(QMargins(leftMargin, 20, 20, 20));
 
     // X-axis is the % frequency. We're already showing percentages on the bar, so we just display 0/50/100%
```
GriffinRichards: This comment isn't accurate anymore

cawtds:
The charts for encounter tables with multiple groups should be fixed.
Bit stumped on the qt5 issue as its compiling locally (with wsl2 ubuntu and qt5). Hopefully the latest commit fixes it :(

GriffinRichards:
Don't worry too much about it, I can re-test / do an actual code review on Monday and fix the build if need be. It's a niche support thing, ideally people will be building with Qt6 and the charts module anyway

cawtds:
> Don't worry too much about it, I can re-test / do an actual code review on Monday and fix the build if need be. It's a niche support thing, ideally people will be building with Qt6 and the charts module anyway

Appreciate it 🙏 
Accidentally broke 6.x, hopefully all builds fixed now...

include/ui/wildmonchart.h
```diff
@@ -74,11 +77,13 @@ public slots:
 
 #else
 
+class Project;
+
```
GriffinRichards: This can be moved outside of `__has_include(<QtCharts>)` so that we don't have to repeat it.

src/ui/wildmonchart.cpp
```diff
@@ -351,7 +393,6 @@ QChart* WildMonChart::createLevelDistributionChart() {
     chart->legend()->setVisible(true);
     chart->legend()->setShowToolTips(true);
     chart->legend()->setAlignment(Qt::AlignBottom);
-    applySpeciesColors(barSets); // Has to happen after theme is set
```
GriffinRichards, line 354: Changing the theme on the level distribution chart is broken now, the colors on the chart remain the same.

src/ui/specieschartview.cpp
```diff
@@ -0,0 +1,102 @@
...
+    for (const QString &name : species)
+    {
+        m_maxTextWidth = std::max(m_maxTextWidth, fm.horizontalAdvance(name));
+
+        QPixmap icon = m_project->getSpeciesIcon(speciesPrefix + name);
```
GriffinRichards:
```c++
QPixmap icon;
if (m_project) icon = m_project->getSpeciesIcon(speciesPrefix + name);
```

GriffinRichards (commented):
Thanks for making these changes! Just a few nitpicks and minor bugs

src/ui/specieschartview.cpp
cawtds: Chose the longest expansion name as the limit now. Longer names get cutoff.

src/ui/wildmonchart.cpp
cawtds: I removed the apply/saveSpeciesColors functions, since with the current setup its not possible anymore (I think?) to have different colors for bars in the species tab.

GriffinRichards:
Thanks! I think there are a few related improvements that could be made here (the tool tip for an elided species name should show the full name, and the current selection for the `Group` combo box should be the same between the two tabs) but those aren't critical. I'll make separate items for those.

